PR #27, opened 2018-05-17
add rukial.tk

jai-dv:


neauoire:
Very nice site :) But I could not find the webring icon anywhere.

jai-dv:
Adding it right now, problems with css...

PR #37, opened 2018-05-18
add v-os.ca

v-exec:
Webring icon is centered in v-os' footer


neauoire:
So good to have you Vi.

PR #38, opened 2018-05-18
Add travisshears.com to webring

BTBTravis:
Webring icon is located in footer as seen here:
webring

**Keep up the good work!**


neauoire:
If your site is on github I can fix the vertical align issue if you like :)

PR #45, opened 2018-05-19
readded window reload

0101solar:
as it seems that is needed for the directory to reload

oops


0101solar:
@neauoire

neauoire:
np :)

neauoire:
yeah, it's still not working.

neauoire:
Could you try the latest version and tell me if it works for you?

0101solar:
Yeah! It's working right now.

PR #66, opened 2018-07-14
adds npisanti.com

npisanti:
Hi! i loved the idea of an old school webring, i always say that internet was much more meaningful back then so i'd really like to be part of it. By the way i really admire the work of lots of you! 

PS: I already put the webring link on my website, in the last section, without the icon as my homepage is only text.


neauoire:
Welcome aboard :)

npisanti:
thanks 
=D

PR #79, opened 2018-09-13
added tynandebold.com

tynandebold:
location of the webring is in the list of nav links at the top of the page


neauoire:
Welcome aboard :)
Beautiful site btw

PR #87, opened 2018-10-20
Update sites.js

cappyzeb:
http://freepages.rootsweb.com/~craigsplace/genealogy/01/rings/

### DIFF
--- a/scripts/sites.js
+++ b/scripts/sites.js
@@ -64,4 +64,5 @@ const sites = [
   , "http://gytis.co"
   , "https://nomand.co"
   , "http://memoriata.com"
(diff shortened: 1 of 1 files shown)


neauoire:
Hi Craig, that's a pretty interesting website you've got there.
Let me browse through a bit :)

cappyzeb:
OH yeah, I like that part too.

neauoire:
Your url is too long and breaks the page layout, I'll try to find a solution and merge this.

wakest:
looks like they are also at the root of this domain http://www.balunsat.org

PR #88, opened 2018-10-23
Add chad.is

ubuwaits:
I've put the icon in the footer of my site as shown in attached image.

[https://chad.is](https://chad.is)

screen shot 2018-10-23 at 16 49 04

Thanks!


neauoire:
Welcome aboard :)

PR #92, opened 2018-11-27
Added iko.soy

ilyakooo0:


neauoire:
Missing: 

- Submit a Pull Request with the location of the webring icon on your site.

The webring's aim is to share hand-crafted websites such as **diaries, wikis & portfolios**, single pagers with nothing but outgoing links to social platforms will be rejected.

ilyakooo0:
The webring icon is at the bottom of every page

PR #93, opened 2018-11-28
Added iko.soy

ilyakooo0:
The webring icon is at the bottom of every page.
https://iko.soy


neauoire:
Better :) Thank you.
Welcome aboard Илка

PR #97, opened 2018-12-23
added magoz.is to sites

magoz:
I've added https://magoz.is :-)


neauoire:
https://github.com/XXIIVV/webring#join-the-webring

neauoire:
You could make it slightly smaller so it fits with the other 2 👍

magoz:
> You could make it slightly smaller so it fits with the other 2 👍

🤔I think it's the cache. It should look like the screenshot:
screenshot 2018-12-24 at 21 14 02

neauoire:
Ah! yes, it looks good now :)

PR #105, opened 2019-01-24
Update sites.js

soyboysky:
I messed up the formatting so the webring wouldn't load. This should fix it, sorry


neauoire:
Thanks @joshavanier, you're in charge while I'm away :)
I should have been more careful with that merge.

PR #106, opened 2019-01-25
Revised my URL

MoreThanLuck:
Revised URL from Monochromatic (now my photography site) to ianjbattaglia, which is a hub site for all my work.


MoreThanLuck:
Thanks friend. Definitely possible, but I couldn't get it to sit right with the other social links. Maybe I'm just not good enough at finessing CSS. Happy to share the code with anyone if they'd like to help. This was the best solution I could think of in the moment.

PR #108, opened 2019-02-04
Add link to dsdshome (https://dsdshcym.github.io)

dsdshcym:


joshavanier:
I don't see the webring icon anywhere on your site

joshavanier:
It's not on your site. Also I'll look into the FF bug later; thanks for the heads up

PR #112, opened 2019-03-20
Hyperlink to target

Arty2:
Sometimes the redirect may fail. A hyperlink to the target page gives an option for the visitor to continue.


neauoire:
Good idea!

PR #114, opened 2019-03-28
added hex22.org to sites.js

kodedninja:
Sadly my site doesn't contain icons, but I've included a short text and the link above the Dat url. I don't know why didn't I do this before? 😅

https://hex22.org


neauoire:
That still counts :3 Welcome aboard

PR #118, opened 2019-04-30
Adding my site to the ring

XiNNiW:
This is a cool idea! I had fun exploring the other sites so I thought I would add my own.

The webring icon is in the page footer.

Hope I'm doing this right!

Cheers,

David


neauoire:
Welcome aboard :)

PR #127, opened 2019-05-17
Add site to list, address unsafe target issue in README

coyote240:
Added my site to the list.

Addressed unsafe target in link recommendation in README.md per this article:
https://www.jitbit.com/alexblog/256-targetblank---the-most-underestimated-vulnerability-ever/


neauoire:
Good call :)

PR #128, opened 2019-05-22
+ Teknari :-)

teknari:


neauoire:
Welcome aboard, it's an honour to have you :)

joshavanier:
Webring link is missing. I think we should also consider adding warnings for NSFW content.

neauoire:
I'm starting to think it would be nice to have a json object with some extra details, maybe like a title, author name and contact infos. Instead of displaying urls, we could display the site's `title`?

ckipp01:
I would love to have an object with contact info such as an email. If that happened, I could build in an email function to the webring-checker so that if a site is down for more than like a week, it uses to contact info to shoot a message to the person notifying them (just in case they don't know).

neauoire:
hey @ckipp01, how about something like [this](https://github.com/XXIIVV/Webring/blob/master/scripts/sites.js#L7)?

ckipp01:
Yea, that would be totally fine!

teknari:
Sorry about the Webring link. Will get right on that.

teknari:
Link is up.

neauoire:
Perfect! Thanks :)

PR #134, opened 2019-05-23
Update xvw.github.io

xvw:
I've add my feed, but it is an Atom feed and not an RSS feed, no problem with that ?


neauoire:
Nha it's fine.

PR #135, opened 2019-05-24
Add CLiC site to sites.js :)

F-3r:


F-3r:
The link to the Webring is in the middle of the page and accessible through the links at the top of the page

PR #141, opened 2019-06-30
Added link to download OPML of all RSS feeds.

jrc03c:
I have no idea if this interests you or not, but I made a link to download an OPML file of all of the RSS feeds in the webring. If you don't want to use it, you won't hurt my feelings. :)


neauoire:
THIS IS BRILLIANT

PR #147, opened 2019-07-02
Add amorris

AAorris:


neauoire:
Welcome : )

AAorris:
Thank you! :)

neauoire:
You're missing an `author` parameter so you won't show up in the feed.

AAorris:
Just noticed that. One sec.

neauoire:
You need to use a tab, or 3 spaces between date and body

neauoire:
that worked!

PR #151, opened 2019-07-02
Add hallway feed (twtxt) for Stephen

setphen:


setphen:
I realized that a gist is not ideal for 2 reasons:

1. Link is super long, hard to `@`
2. The link is to a specific commit to the gist, and will not update as I update the gist itself.

i will create another PR with a better feed link.

neauoire:
please, ideally it would have your site url in it.

PR #153, opened 2019-07-03
added wake.st/twtxt.txt

wakest:
I just added a .txt file manually to my github repo and copied the most recent timestamp from the hallway, I wonder if that will mess anything up


neauoire:
That should be fine, let's find out.

PR #167, opened 2019-07-06
Update sites.js

szymonkaliski:


kodedninja:
Welcome to the Webring! 

Fyi: generally we link the Webring icon to `https://webring.xxiivv.com/#random` to support exploring random sites and remove the importance of the list (at least that's why I do it), but I guess it was intentional.

PR #172, opened 2019-07-12
Update sites.js

serocell:

PR #173, opened 2019-07-12
fix missing comma

ckipp01:


ckipp01:
Never mind #172 will fix this as well when it's merged.

PR #186, opened 2019-07-21
Update sites.js

Imperial-Industries:


scripts/sites.js
```diff
@@ -95,4 +95,5 @@ const sites = [
   { url: 'https://kor.nz', title: 'kor', type: 'wiki', author: 'kormyen', contact: 'h@kor.nz' },
   { url: 'https://lublin.se', author: 'quite', contact: 'quite@hack.org', feed: 'https://lublin.se/twtxt.txt' },
   { url: 'https://zanneth.com', title: 'zanneth.com', author: 'zanneth', type: 'blog', contact: 'root@zanneth.com' }
```
ckipp01: You're missing a comma on this line which will break the webring.

kodedninja:
I don't think this website fits next to all the members of the Webring, which are mostly personal wikis, portfolios and blogs. It is by far different even from the currently marginal sites.

ckipp01:
@Imperial-Industries after a short discussion with another maintainer we feel that your site may not be the best fit for the webring. While we appreciate your interest, as the readme mentions, this webring is primarily for personal wikis, diaries, and blogs to both inspire others and share traffic among the likes. Your site doesn't seem to mesh well with the current sites, and also tends to lean a bit more towards being a business website where you are selling services. This isn't the point of the webring.

Cheers.

PR #187, opened 2019-07-21
switch user regex to be twtxt compliant

### LINKED ISSUE #171: Implement twtxt native mentions support
Halp @Bradshaw

ckipp01:
So I gave this a go today. I'm not sure of a better way to do it. For now, I just grab everything between the @< and > and wrap the user span around it. I wanted to capture the user but having a @<test.test.com/test.txt> is a valid mention in twtxt land. Also, if people are using the same feed in the hallway as their twtxt feed, then they may mention people not in the hallway, meaning we have no way to know their user. Therefore, I just opted for this.

This will close #171 

hallway

### DIFF
--- a/scripts/hallway.js
+++ b/scripts/hallway.js
@@ -1,7 +1,7 @@
 'use strict'
 
 const re_channel = /(\s|^)\/([a-zA-Z0-9]+)/g
-const re_user = /((\s|^)@&lt;[a-zA-Z0-9./: ]*&gt;)/g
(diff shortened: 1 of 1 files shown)


ckipp01:
Hold off on reviewing this for a sec, I have something else to add to it.

kodedninja:
Hmm, personally I would prefer seeing usernames only, I thought that was the plan, but I understand your arguments too.

ckipp01:
Yea, @kodedninja I totally get what you mean. Let me work on this a bit more. I should be able to get it that if the mention is someone in the hallway it replaces it with @username, and if not, then it will display the entire mention.

PR #200, opened 2019-07-24
Parse glossary from site wiki

AAorris:
Posting up mostly for reference/discussion rather than an intent to merge

```
[x] Code can be content-type aware and expect a glossary link on a wiki homepage.
[x] Example: hover over the new wiki link to see alt-text of a random wiki's glossary.
[ ] Actually parse the ndtl and do something useful to it.
[ ] Allow a content type of `application/json` which simply skips indental parsing.
```

### DIFF
--- a/index.html
+++ b/index.html
@@ -8,6 +8,7 @@
   <script src="scripts/portal.js"></script>
   <script src="scripts/sites.js"></script>
   <script src="scripts/opml.js"></script>
+  <script src="scripts/wiki.js"></script>
   <title>Webring</title>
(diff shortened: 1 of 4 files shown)


neauoire:
I can't test it right now, but tomorrow I'm onto this ✊

AAorris:
Me too - just posting the idea to discuss before I head off to bed!

AAorris:
This could use more thought before being considered. It's - at best - a quick reference on how multiple content types might be handled

Issue #194, opened 2019-07-21
No webring logo and copy of tumblr


ckipp01:
Hey @streeetlamp! Just wanted to reach out about your site now in the webring. We're currently doing a check through all the sites and noticed your site wasn't really in line with multiple of our rules. If we understand it correctly, it's just a portal to your tumblr since it seems to just be posting from your tumblr. Secondly, there is no webring logo. Also, you have a lot of trackers/adds going on in your site, which isn't really in line with the values of the webring.

I'll leave this up for a few days before removing the site just in case you're currently moving some things around or want to replace this site with another you may have.

neauoire:
Actually, this is an instant removal from me, no link-back websites don't get a week's time to correct.

streeetlamp:
Hi, very sorry for missing this notification. I have updated my site and submitted a PR. Please let me know if I have made any mistakes or overlooked something. Thanks

PR #205, opened 2019-07-25
Update sites.js to be in compliance

streeetlamp:
Webring link is in the header of site, thanks. 
Sorry I did not see the notification of my removal from the webring, I have now made updates. Thank you


ckipp01:
Hey @streeetlamp, thanks for addressing the webring icon.

Regardig #194 , one of the other things we had mentioned was it looked like your site may just be a copy of your tumblr activity, which would be in violation of the [webring criteria](https://github.com/XXIIVV/Webring#webring-criteria). Could you give us a bit more info on how the content on your site is curated? Does it just get generated from your Tumblr activity, or another way?

kodedninja:
@ckipp01 as discussed earlier, I totally agree with you, but I'm also curious about the intent of the creator.

neauoire:
I don't think it qualifies as a hand-crafted website, or a place to showcase the creator's audio/visual work.

streeetlamp:
Once again sorry for the delay in responding. My site only uses tumblr as a host for content I curate for many reasons like particular interests and a shifting mood board that has been curated from content I have come across or created. I have designed the site and written the markup of the site. I could move content to a different platform if that changes opinions for some reason.

PR #207, opened 2019-07-25
Added twtxt link

et0and:
Added myself a twtxt link, I'm sorry that I forgot to do this earlier!


ckipp01:
Hey @et0and it looks like there is a CORS issue with pulling your twtxt file from your server, which why you aren't seeing your posts. You can see an example of how to fix this in the readme [here](https://github.com/XXIIVV/Webring#joining-the-hallway).

Let us know if you need any help!

PR #218, opened 2019-07-30
Allow to join the Hallway without Website

xvw:
I don't know if it follow your idea, but I see people which want to join the Hallway, but with a Work In Progress website.


neauoire:
The webring is to encourage people to build websites before taking part in the hallway. 
Applicants need both :)

xvw:
Get your point (I suspected it a bit, hence the "I don't know if it follow your idea").
Cheers!

PR #219, opened 2019-07-30
Removing entry for Gaeel / spaceshipsin.space

Bradshaw:
I no longer want to be a part of webring, so am removing my entry

### DIFF
--- a/scripts/sites.js
+++ b/scripts/sites.js
@@ -21,7 +21,6 @@ const sites = [
   { url: 'https://felipecortez.net' },
   { url: 'http://luminghao.com' },
   { url: 'https://turelio.github.io' },
-  { url: 'https://spaceshipsin.space', title: 'spaceshipsin.space', type: 'hybrid', author: 'gaeel', contact: 'gaeel@spaceshipsin.space', feed: 'https://spaceshipsin.space/twtxts/hallway.txt' },
   { url: 'https://log.lectronice.com', title: 'lectrolog.', type: 'blog', author: 'lectronice', contact: 'i.love.spam@lectronice.com', feed: 'https://lectronice.com/hallway/twtxt.txt' },
   { url: 'https://craze.co.uk' },
   { url: 'https://shaneckel.com' },


ckipp01:
Sorry to see you go, Gaeel. If you change your mind, feel free to send pr, and we'd gladly welcome you back.

Bradshaw:
I'm getting a few questions about this, which is completely understandable
I left the Merveilles Slack last night and am leaving the webring too, but will stay on the merveilles.town instance for now, so I'm not completely running away   
   
I'm mostly just trying to understand what Merveilles actually represents, to me, to the world, and to other Merveillans, in part with regards to a merveilles.town conversation about Merveilles' values last week   
   
I'm still having trouble formulating what my worries are, but I'd like to encourage you all to look at what values Merveilles clearly puts forwards, _rationality, technology, self-reliance, quantifiability, analysis_, and question what place there is there for empathy and care.
I'm am absolutely not saying that Merveillans lack empathy or care, in fact everyone I've spoken to within Merveilles seem to me to be empathetic and caring. I'm just worried that the aesthetics in our community might have a tendency to one day go against the ideals that I value the most.

ckipp01:
Thanks for the comments @Bradshaw.

I'll tag @neauoire and @kodedninja just make sure they take a gander at this as well.

I don't have much to say at all because I want to think on what you said for a while. I do however just want to mention that I think what your saying is valid and also that it's a bit hard to approach this, because the three concrete groups you mentioned 1) Merveilles Slack, 2) Merveilles.town, and 3) this webring are all completely different entities in my eyes that serve very different purposes for different reasons. For me personally, I actually view this Webring as the XXIIVV webring, primarily something that @neauoire started completely separate from Merveilles. I may be wrong on that, but since this isn't in that "groups" repos, I don't view it as such. I think that's an important distinction.

 As for the Slack group and Merveilles.town, I think a large discussion could be had on that, but probably in a better place than here.

Again, thanks for your thoughts, Gaeel.

neauoire:
Yeah, the webring is not merveilles. But that's absolutely fine, I understand. Merveilles tends to be a very cold place. 

Is it possible I might have perhaps missed a conversation where someone lacked tact or was rude or something, I feel I might be missing context here? If so, please let me know aliceffekt@gmail.com

PR #225, opened 2019-07-30
Update sites.js

electricgecko:
After losing the webring icon with the latest iteration of my site, re-added it to the contact/about page (http://electricgecko.de/kontakt), among other active accounts.


neauoire:
Thanks :)

PR #232, opened 2019-08-03
Filter out messages with external references

ckipp01:
This will do what we mentioned on the hallway regarding filtering out messages that reference individuals that aren't in the hallway. I think this will be a good idea for people that are using twtxt and not just following people in the hallway. This way messages can't be targeted at people not in the hallway, and then those messages just get filtered out.

I also did a bit of cleanup on various things.


neauoire:
Thanks for cleaning up the underscores :)

PR #238, opened 2019-08-05
[sites] fix duplicated `feed:` property

dym-sh:
...forgot to make another branch

also fixed the "undefined" class on older sites w/o `type:`
and unified output for whatever `<type>` might be

### DIFF
--- a/scripts/portal.js
+++ b/scripts/portal.js
@@ -18,7 +18,7 @@ function Portal (sites) {
   function _directory (sites) {
     return `
     <ul>${sites.reduce( (acc, site, id) =>
-          `${acc}<li class='${site.type || ''}' id='${id}'><span class='counter'>${id})</span><a href='${site.url}'>${_name(site)}</a></li>`
+          `${acc}<li data-type='${site.type || ''}' id='${id}'><span class='counter'>${id})</span><a href='${site.url}'>${_name(site)}</a></li>`
     , '')}</ul>\n${_readme()}${_buttons()}`
   }
 
--- a/scripts/portal.js
+++ b/scripts/portal.js
@@ -18,7 +18,7 @@ function Portal (sites) {
   function _directory (sites) {
     return `
     <ul>${sites.reduce( (acc, site, id) =>
-          `${acc}<li data-type='${site.type || ''}' id='${id}'><a href='${site.url}'>${_name(site)}</a></li>`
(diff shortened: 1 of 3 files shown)


kodedninja:
I like the changes, but as you touched @neauoire's entry also, I would like him to check it too.

scripts/portal.js
```diff
@@ -17,11 +17,13 @@ function Portal (sites) {
 
   function _directory (sites) {
     return `
-    <ul>${sites.reduce((acc, site, id) => { return `${acc}<li class='${site.type}'><span class='counter'>${id})</span><a href='${site.url}'>${_name(site)}</a></li>` }, '')}</ul>\n${_readme()}${_buttons()}`
+    <ul>${sites.reduce( (acc, site, id) =>
+          `${acc}<li data-type='${site.type || ''}' id='${id}'><span class='counter'>${id})</span><a href='${site.url}'>${_name(site)}</a></li>`
```
ckipp01: Is `id='${id}'` added in here for a specific reason?

ckipp01:
> I like the changes, but as you touched @neauoire's entry also, I would like him to check it too.

I actually changed this in my pull request as well that was just merged. So that shouldn't be an issue, but I did have a question on the added id.

dym-sh:
i was thinking about adding numerations into some css `::before` instead of html, i'll add to the PR in a bit

ckipp01:
Thanks for this @dym-sh!

scripts/portal.js
```diff
@@ -18,7 +18,7 @@ function Portal (sites) {
   function _directory (sites) {
     return `
     <ul>${sites.reduce( (acc, site, id) =>
-          `${acc}<li data-type='${site.type || ''}' id='${id}'><a href='${site.url}'>${_name(site)}</a></li>`
+          `${acc}<li ${site.type ? `data-type="${site.type}"` : ''} id='${id}'><a href='${site.url}'>${_name(site)}</a></li>`
```
dym-sh: > data-type="${site.type}"
should i add escaping of possible double quotes in site-types? like:
``` js
`data-type="${site.type.replace(/"/g, '&quot;')}"`
```

ckipp01: My initial opinion is no since none of the site types should have a quote in them.

PR #245, opened 2019-08-05
Update info for Forida, add hallways entry

mboxed:
Hey everyone,

Just a quick change, seeing that hallways has happened :)


neauoire:
Could you also add a way to contact you?

mboxed:
Absolutely! Sorry about that, forgot that entry :)

PR #247, opened 2019-08-07
[wiki] fix sidebar navigation constraints

dym-sh:
general problem i have is this site w/o zooming-in is unreadable:
Screenshot--20190807-151126

but with zoom, sidebar list is just cut off:
Screenshot--20190807-152756

now it has a vertical scrollbar.


neauoire:
thanks :)

Issue #242, opened 2019-08-05
Use more semantic html


ckipp01:
For example we don't use `<footer>` tags and such, which doesn't seem like a huge deal, but it helps with accessibility for anyone that may be using a screen reader or anything. Just a quick run through to replace some of the p tags or wrap some of them in the alternative semantic elements would be a nice improvement.

PR #249, opened 2019-08-08
[wiki] fixed sidebar and navigation

### LINKED ISSUE #241: Don't cut off categories mid word
image

We should probably get a better way to do this. Either change the way the sidebar is done to better accomodate longer phrases, or cut off at a full word or at least have a ... to end a word or something so it doesn't just stop at the end of a word.

### LINKED ISSUE #248: Scrollbar causing odd line breaks
#247 added in a scrollbar for the sidebar, but with the current layout that is causing a line break and odd display.

image

dym-sh:
closes #241
closes #248

partial #242

opening, back/ff history, and manually refreshing a page with \#HASH now also loads the same contents
downside is `%20` and whatnot in the url's hash, will fix later

_sidenote_, why use upper case in the internal representation, if it always shows in lower case everywhere else?


ckipp01:
Also regarding your questions about uppercase in the data structure and then lowercase in the display. I really don't know actually. That's a question for @neauoire.

PR #256, opened 2019-08-09
Feature: Wiki header showing the page name

ljcooke:
This adds a `<header>` to the wiki which shows the full name of the page you're on.

### DIFF
--- a/links/main.css
+++ b/links/main.css
@@ -83,6 +83,7 @@ body #portal ul li a:hover { text-decoration: underline; }
 /* Wiki */
 
 body #wiki { background: #ccc; color:#333; }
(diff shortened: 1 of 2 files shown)


ckipp01:
Hey @ljcooke thanks for doing these! I agree, these are good changes. I have a huge pr in #258 that changes a lot of how the wiki stuff is done, and it also includes the same changes you made basically, so let's see what happens with that first. If that gets merged it won't be necessary to merge this, if not, then we'll merge this in. I just want to a void a merge conflict if I merge this in right now.

ljcooke:
@ckipp01 Nice work! Just had a look at your PR – I like the approach you've taken. Makes sense to avoid a merge conflict with this one. 👍

PR #260, opened 2019-08-10
Feature: Add a progress indicator to the wiki

ljcooke:
This shows a progress indicator in the footer on the Wiki, so you can tell when everything's loaded. I've tried to make it general enough that the same code could be run on the Hallway.

Screenshot: 88% loaded

Screenshot: 100% loaded

### DIFF
--- a/links/main.css
+++ b/links/main.css
@@ -77,8 +77,10 @@ body #wiki #footer { padding-left: 60px; background-image: url(../icon.white.svg
 /* progress indicator */
 body #progress { color: #777; }
 body #progress .busy { cursor: wait; }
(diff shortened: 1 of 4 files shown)


ljcooke:
I've been testing in Firefox, but it looks odd in Chrome & Safari. Gonna test some more.

ljcooke:
Fixed.

It seems Webkit doesn't like it when rules are shared between `-webkit` and `-moz` selectors. So something like this:

```css
::-webkit-thing, ::-moz-thing { background: #777 }
```

…must be changed to:

```css
::-webkit-thing { background: #777 }
::-moz-thing { background: #777 }
```

🙃

PR #262, opened 2019-08-11
add in more issue templates

ckipp01:


neauoire:
Good idea to split them like that.

PR #278, opened 2019-08-12
Feature/merge same definitions

### LINKED ISSUE #263: Join duplicate definitions
**Please describe the issue in the wiki:**
Currently if two definitions under a category are exactly the same, they are both displayed like so

image

I think it would be nicer to just display the definition once and then show both authors. Something like `- author1 / author2`

ckipp01:
closes #263 

Before:

image

After:

image

This will only happen when the two definitions are identical.


neauoire:
you
are
a
machine

PR #288, opened 2019-08-14
Add dotcomboom.somnolescent.net

dotcomboom:
👋 🐍
Webring icon is in the footer:
image


ckipp01:
Also, you'll have to add some headers in to deal with CORS in order for us to read your twtxt file for the hallway.

image

PR #295, opened 2019-08-17
change around the list entry to make unique to make sure it's counted

### LINKED ISSUE #294: Look at how we count entries
**Please describe the issue in the wiki:**

This is more of just a reminder to myself, but we should take a look at how we are counting entries and displaying the number when you are on /home. If a user adds in a new list, that list isn't actually counted as an entry.

Change the counting around to also count a list entry as an entry to get a more accurate count.

ckipp01:
closes #294


ckipp01:
This was a really small change. I hope it's ok that I just went ahead and merged it.

Issue #39, opened 2018-05-18
Move commas to front of line for cleaner commits.


BTBTravis:
What about moving the commas to the beginning of the line to make the adding new site commits cleaner? This would make pull request of folks adding a site will only be a single line addition instead an edit of the previous line and an addition.

ex:
``` 
var sites = [
   "https://wiki.xxiivv.com/"
   , "http://estevancarlos.com/"
   , "https://electro.pizza/"
   , "http://xla.wiki/"
   , "https://monochromatic.co/"
   , "https://joshavanier.github.io/"
   , "http://kaemura.com/"
...
```

neauoire:
that's an amazing idea. I'll make the change today.

PR #304, opened 2019-09-08
make sites.js human-readable

dym-sh:
this PR re-implementes "commas-first" structre, first proposed and approved in #39
( https://github.com/XXIIVV/Webring/commit/3a75ddaf60ff22c567a4d84e958e912dcd1fe343#diff-eacf331f0ffc35d4b482f1d15a887d3b )

re-formated sites.js, to make it human-readable;
this will also make future editing (like adding site-types) simplier and clearer.

order and functionality stayed unchanged.

### DIFF
--- a/scripts/sites.js
+++ b/scripts/sites.js
@@ -4,603 +4,603 @@
 // protocol://url.domain.ext
 
 const sites = [
-{
-  url: 'https://wiki.xxiivv.com',
-  title: 'xxiivv',
-  type: 'wiki',
-  author: 'neauoire',
-  contact: 'aliceffekt@gmail.com',
-  rss: 'https://wiki.xxiivv.com/links/rss.xml',
-  feed: 'https://wiki.xxiivv.com/twtxt.txt',
-  wiki: 'https://wiki.xxiivv.com/scripts/database/glossary.ndtl',
-},
-
-{
-  url: 'http://estevancarlos.com'
-},
-
-{
-  url: 'https://electro.pizza',
-  title: 'electro pizza',
-  type: 'blog',
-  author: 'rho',
-  rss: 'https://electro.pizza/feed.xml',
-  feed: 'https://electro.pizza/twtxt.txt',
-},
-
-{
-  url: 'https://avanier.now.sh',
-  title: 'Arachne',
-  type: 'wiki',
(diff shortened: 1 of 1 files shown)


ckipp01:
My personal preference is that we shouldn't do this. I know it used to be this way, and I really disliked it since it doesn't follow a standard "linting" practice. The way it is now, it can be run through Standard easily checked for single quotes, space between object brackets, and trailing commas. If we implement this, we can no longer do this.

If we are going to attempt to make this more easily readable, then I recommend that we put each key-value pair on its own line just like you did, but not with a leading comma unless we enforce that throughout the project, which I'd be against.

I'm much more in favor of this:
```
{ 
  url: 'https://electro.pizza',
  title: 'electro pizza',
  type: 'blog',
  author: 'rho',
  rss: 'https://electro.pizza/feed.xml',
  feed: 'https://electro.pizza/twtxt.txt'
}
```
rather than this:
```
, { url: 'https://electro.pizza'
    , title: 'electro pizza'
    , type: 'blog'
    , author: 'rho'
    , rss: 'https://electro.pizza/feed.xml'
    , feed: 'https://electro.pizza/twtxt.txt'
    }
```

neauoire:
I'm pretty much ambivalent between the current format and a standard friendly unfolded display for readability.

```
{ 
  url: 'https://electro.pizza',
  title: 'electro pizza',
  type: 'blog',
  author: 'rho',
  rss: 'https://electro.pizza/feed.xml',
  feed: 'https://electro.pizza/twtxt.txt'
}
```

dym-sh:
ok, now it's comma-lst

dym-sh:
yeah, that's intentional, and officially acknowleged usecase: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Trailing_commas

dym-sh:
ok, no trailing commas

ckipp01:
> yeah, that's intentional, and officially acknowleged usecase: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Trailing_commas

True, but the project is using Standard which [doesn't allow comma-dangle](https://standardjs.com/rules.html). Looks good though. Thanks!

PR #305, opened 2019-09-08
[portal] added toggles for site-types

### LINKED ISSUE #213: Add sidebar, filter by site type

dym-sh:
closes #213, probably

webring-portal--types

### DIFF
--- a/index.html
+++ b/index.html
@@ -15,15 +15,6 @@
     const portal = new Portal(sites);
     portal.install(document.body);
     portal.start();
-
-    function toggleType(e) {
-      const navLink = e.target
-      const selectedType = navLink.dataset.type
-      document.querySelector(`nav a.current`).classList.remove('current')
-      navLink.classList.add('current')
-      document.querySelectorAll(`main li`).forEach(el => el.style.display = 'block')
-      if(selectedType) { document.querySelectorAll(`main li:not([data-type='${selectedType}'])`).forEach(el => el.style.display = 'none') }
-    }
   </script>
 </body>
 </html>
--- a/index.html
+++ b/index.html
@@ -16,22 +16,14 @@
     portal.install(document.body);
     portal.start();
 
-    function toggleType( e )
-      { const navLink = e.target
-            , selectedType = navLink.dataset.type
-
-        document.querySelector( `nav a.current` )
-                .classList.toggle( 'current' )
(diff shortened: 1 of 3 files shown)


neauoire:
Please follow the coding style of the rest of the project.

dym-sh:
:+1:

ckipp01:
One other thing to think about (I'm also not able to test this at the moment to see how it responds), but it'd be cool if when someone filtered by type, it condensed the list down rather than just hiding the non-matching site types. My main reason for saying this is that now on mobile (I believe), there is going to be huge gaps since all of the sites are going to be in a single column. I think that will be a bad user experience on mobile.

dym-sh:
now uses `display:none` to hide

test under https://dym-sh.github.io/Webring/

webring-portal--types-v2

index.html
```diff
@@ -15,6 +15,15 @@
     const portal = new Portal(sites);
     portal.install(document.body);
     portal.start();
+
+    function toggleType(e) {
```
ckipp01: Are you able to move this out of index.hml and into portal.js? Then, just change the onClick handers to be `portal.toggleType(event)`.

scripts/portal.js
```diff
@@ -16,8 +16,10 @@ function Portal (sites) {
   }
 
   function _directory (sites) {
+    const siteTypesArray = [...new Set(sites.map(site => site.type).filter(Boolean))]
+    const siteTypes = siteTypesArray.reduce((output, siteType) => `${output}<a data-type='${siteType}' href='#' onclick='event.preventDefault(); toggleType(event)'>&lt;${siteType}&gt;</a>`, `<a class='current' href='#' onclick='event.preventDefault(); toggleType(event)'>&lt;all&gt;</a>`)
```
ckipp01: I don't believe the `event.preventDefault()` is actually doing anything here is it?

dym-sh: it prevents the `#` from href to apply to url

ckipp01: Ahhh, duh, of course. Thanks!

index.html
dym-sh: sure

ckipp01: 👍

ckipp01:
Looks good, thanks @dym-sh!

PR #312, opened 2019-09-14
Site Languages

### LINKED ISSUE #282: How to handle languages
Created this as a follow-up and continuation of this conversation -> https://merveilles.town/@neauoire/102606176450333603

I'll outline basically what @joshavanier mentioned since I think that may be a good technical solution, and then we can hammer out some of the specific details of what we want to support as well.

To take what Josh said a bit further, instead of just using `navigator.language` we could use `navigator.languages` to get the top 2-3 most used language settings for the user and that can be passed along with the user hits `random`. When the hit random we can just filter out the sites that aren't listed in the users top languages to avoid them getting a site in a language they never browse, but also allowing for example an English/French speaker to get both English and French sites.

I think this could also relate to #213 and we could put in some sort of filter system that you can filter the sites by type and/or by language.

Unless of course the decision is to not support non-English sites. That's ultimately your call @neauoire

ckipp01:
closes #282 

This was pretty fun to implement, and I think it will work pretty well. This implements 2 different things:

1) When a user hits `#random` or is directed there, the langs that they have used are grabbed from [navigator.languages](https://developer.mozilla.org/en-US/docs/Web/API/NavigatorLanguage/languages) which will return an array of their browsers preferred languages. Then if a site contains a language code that matches any of their preferred languages, it will be a possibility. If not, it will be filtered out. However, they can still see that site if they are looking at the listing of course.

2) Regarding the listing, you can now also see the site langs next to the site type. I've included a screenshot below to illustrate. I just went through all the sites and did my best to include the language represented there. I only included a second language if there was at least like a full blog entry or something in that language.

Screenshot 2019-09-14 at 20 41 38

### DIFF
--- a/scripts/portal.js
+++ b/scripts/portal.js
@@ -13,8 +13,7 @@ function Portal (sites) {
   function _directory (sites) {
     const siteTypesArray = [...new Set(sites.map(site => site.type).filter(Boolean))]
     const siteLangsArray = [...new Set(sites
-      .map(site => site.langs)
-      .reduce((flattened, langs) => flattened.concat(langs), [])
+      .reduce((flattened, site) => flattened.concat(site.langs), [])
       .sort())]
 
     const siteTypes = siteTypesArray.reduce((output, siteType) =>
@@ -39,8 +38,7 @@ function Portal (sites) {
   }
 
   function _lang (site) {
-    const langs = site.langs.reduce((langs, lang) => `${langs} ${lang}`, '')
-    return `data-lang="${langs.trim()}"`
+    return `data-lang="${site.langs.join(' ').trim()}"`
   }
 
   function _redirectView (site) {
@@ -73,8 +71,7 @@ function Portal (sites) {
   }
 
   this.sitesMatchingLangs = this.sites
-    .filter(site => site.langs)
-    .filter(site => site.langs.some(lang => navigator.languages.includes(lang)))
+    .filter(site => site.langs && site.langs.some(lang => navigator.languages.includes(lang)))
 
   this.next = function () {
     const hash = window.location.hash.replace('#', '').trim()
--- a/scripts/portal.js
+++ b/scripts/portal.js
@@ -7,22 +7,27 @@ function Portal (sites) {
 
   // Templates
 
-  function _readme () {
-    return '<p class="readme">This webring is an attempt to inspire artists & developers to build their own website and share traffic among each other. The ring welcomes personalized websites such as <b>diaries, wikis & portfolios</b>.</p><p>To add yourself to the ring, submit a <a href="https://github.com/XXIIVV/webring/edit/master/scripts/sites.js" target="_blank">Pull Request</a>.<br />If you found a broken link, please <a href="https://github.com/XXIIVV/webring/issues/new" target="_blank">report it</a>.</p>'
-  }
-
-  function _buttons () {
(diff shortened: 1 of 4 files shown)


ckipp01:
@neauoire is it possible to get this looked at before merging anything else in. Since this touches every entry in sites.js it ends up creating a merge conflict every time we add/edit an entry, and I'm continually having to rebase.

scripts/portal.js
```diff
@@ -7,33 +7,43 @@ function Portal (sites) {
...
     return site.title || `${site.url.split('//')[1]}`
   }
 
   function _type (site) {
-    return site.type ? `data-type="${site.type}"` : ''
+    return site.type ? `data-type="${site.type} "` : ''
+  }
+
+  function _lang (site) {
+    const langs = site.langs.reduce((langs, lang) => `${langs} ${lang}`, '')
```
xvw: I think that you can use, here, https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Objets_globaux/Array/join

```javascript
const langs = site.langs.join(' ')
```

scripts/portal.js
```diff
@@ -7,33 +7,43 @@ function Portal (sites) {
...
-  }
+  const _readme = '<p class="readme">This webring is an attempt to inspire artists & developers to build their own website and share traffic among each other. The ring welcomes personalized websites such as <b>diaries, wikis & portfolios</b>.</p><p>To add yourself to the ring, submit a <a href="https://github.com/XXIIVV/webring/edit/master/scripts/sites.js" target="_blank">Pull Request</a>.<br />If you found a broken link, please <a href="https://github.com/XXIIVV/webring/issues/new" target="_blank">report it</a>.</p>'
+  const _buttons = '<p class="buttons"><a href="#random" onClick="portal.reload()">Random</a> | <a href="https://github.com/XXIIVV/webring">Information</a> <a id="icon" href="#random" onClick="portal.reload()"></a> | <a href="hallway.html">The Hallway</a> | <a href="wiki.html">The Wiki</a> | <a id="opml">OPML</a></p>'
 
   function _directory (sites) {
     const siteTypesArray = [...new Set(sites.map(site => site.type).filter(Boolean))]
+    const siteLangsArray = [...new Set(sites
+      .map(site => site.langs)
+      .reduce((flattened, langs) => flattened.concat(langs), [])
```
xvw: map + reduce is, in fact, reduce :) 

```javascript
siteLangsArray = [...new Set(sites
      .reduce((flattened, site) => flattened.concat(site.langs), [])
      .sort()]
```

scripts/portal.js
```diff
@@ -62,30 +72,48 @@ function Portal (sites) {
     setTimeout(() => { window.location.href = target }, 3000)
   }
 
-  this.locate = function () {
+  this.sitesMatchingLangs = this.sites
+    .filter(site => site.langs)
+    .filter(site => site.langs.some(lang => navigator.languages.includes(lang)))
```
xvw: Wdyt?
```javascript
.filter(site => site.langs && site.langs.some(lang => navigator.languages.includes(lang)))
```

scripts/portal.js
ckipp01: o duh, of course, way simpler.

scripts/portal.js
ckipp01: Indeed. Fixed.

scripts/portal.js
ckipp01: Another good catch. Fixed.

ckipp01:
Thanks a lot for the feedback @xvw!

PR #321, opened 2019-09-21
small fixes in readme and url trimming

dym-sh:
- unexpected single qoutation mark in readme
- remove final slashes from sites-entries w/o site-names (i think i saw a rule or smth related to that before, but it looks better now anyway)
- handle urls with final slashes from now on


neauoire:
Thanks :)

PR #322, opened 2019-09-21
re-worded site-entry spec

dym-sh:
i think it is more clear about which parameters are needed for what parts


neauoire:
I agree.

PR #323, opened 2019-09-24
Adding myself to the webring

0xroy:
Webring icon is located in the navbar

### DIFF
--- a/scripts/sites.js
+++ b/scripts/sites.js
@@ -744,6 +744,17 @@ const sites = [
     rss: 'https://icyphox.sh/blog/feed.xml',
     type: 'hybrid',
     url: 'https://icyphox.sh'
+  },
+
+  {
+    author: 'roy niang',
+    contact: 'roy@royniang.com',
+    title:  'roy niang',
+    url: 'http://royniang.com',
+    type: 'portfolio',
+    feed: 'https://royniang.com/updates.txt'
+    wiki: 'https://royniang.com/usefulnotes.ndlt',
+    langs: ['en', 'fr'],
   }
 
 ]
--- a/scripts/sites.js
+++ b/scripts/sites.js
@@ -750,7 +750,7 @@ const sites = [
     author: 'roy niang',
     contact: 'roy@royniang.com',
     feed: 'https://royniang.com/updates.txt',
-    langs: ['en', 'fr'],
+    langs: ['en'],
     title:  'roy niang',
     type: 'portfolio',
     url: 'http://royniang.com',
--- a/scripts/sites.js
+++ b/scripts/sites.js
@@ -749,12 +749,12 @@ const sites = [
   {
     author: 'roy niang',
     contact: 'roy@royniang.com',
-    title:  'roy niang',
-    url: 'http://royniang.com',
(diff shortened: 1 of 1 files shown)


0xroy:
I also missed a comma, I just woke up ahah

Cheers

ckipp01:
Hey, @0xroy another question! I noticed you had `fr` listed under the languages, but I don't see any French on your site. This is a new addition to the webring, and maybe not clear (I'll update the readme if that's the case), but the language key is for languages that are represented on your site. If you don't have any French on your site, I'd remove that from the language array. That way people who are filtering to find French sites won't include yours, which doesn't have French. If I missed the French portion of your site, ignore this!

0xroy:
I do speak French, but I thought that the parameter was not limited to the content and included contact, feed, wiki, etc.

I also included French because, even if unused, it’s still a networking parameter.

If it is limited to the content, I’ll remove it :)

ckipp01:
I'd say yea, remove it for now. That field is only being used for the Portal portion of the website at the moment, which represents the content on your site. Sorry for the confusion on this!

0xroy:
Removed :)

Cheers!

ckipp01:
Alright, a couple more things actually @0xroy. You Wiki and Hallway entries aren't showing up yet because there is a [CORS](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS) issue. You'll need to allow this domain to access those resources. Let me know if you have any questions about this.

0xroy:
Mmmh, this is on the server level right? I’ll ask if it’s possible

ckipp01:
Correct.

The files needs to be accessible via CORS, so for example if you're self-hosting, you need to make sure to allow the `webring.xxiivv.com` origin.

Here's an example of how to do so with Nginx:

```
location / {
    index index.html;
    if ($request_method = 'GET') {
        add_header 'Access-Control-Allow-Origin' 'https://webring.xxiivv.com';
        add_header 'Access-Control-Allow-Methods' 'GET';
        add_header 'Access-Control-Allow-Headers' 'DNT,User-Agent,X-Requested-With,If-Modified-Since,Cache-Control,Content-Type,Range';
        add_header 'Access-Control-Expose-Headers' 'Content-Length,Content-Range';
     }
}
```

If you're hosting with someone else, you may need to just search how to do it or see if it's even possible (which it should be).

ckipp01:
O, and one other thing. It looks like your ndtl file can't be found? Your site is returning a 404 when you try to fetch it.

0xroy:
https://github.com/XXIIVV/Webring/pull/324#issue-320713523

Fixed it here, simple file extension error, i’m not yet familiar with this format! `;` is the comment symbol right?

ckipp01:
Cool, I see everything getting pulled in now correctly with the webring-cli. Once the cors stuff gets resolved you should see it show up in the hallway and wiki.

As for ndtl. If you didn't see it, @neauoire has some details [here](https://wiki.xxiivv.com/#indental) which talks a bit about it. Although, now that I look at it, it doesn't mention the `;` being a comment, but yes, it is.

PR #328, opened 2019-10-06
add in styling to accommodate for more sites in the portal

ckipp01:
Screenshot 2019-10-06 at 18 02 55

We are getting to the number of sites that 3 columns aren't working the best as it's causing the screen to have to scroll a bit. This will allow for the sites to be able to keep growing and displaying nicely. Basically, just bump to 4 columns when the screen is wide enough. I also added in a sticky footer, since I think it looks nice.


neauoire:
yup

PR #334, opened 2019-10-18
small updates

ckipp01:
Just a few really small edits:

1) New issues template general to the portal - I realized we didn't have one
2) Moved the reminders of protocol and other style things to the bottom of the sites.js file. I figured it made more sense to have it at the bottom than then top since most people are just editing the bottom of that file, probably never seeing or reading the message at the top.
3) Removed the webring-checker from the readme. The usage has spiked on that the last couple months causing my zeit bill to go up quite noticeable. The usage is pretty much all on the rss viewer, but that can be provided by the cli. So I'd rather people use that.

### DIFF
--- a/.github/ISSUE_TEMPLATE/portal-issue.md
+++ b/.github/ISSUE_TEMPLATE/portal-issue.md
@@ -1,0 +1,10 @@
+---
+name: Portal issue
+about: Report an issue with the Portal
+title: ''
+labels: portal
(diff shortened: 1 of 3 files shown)


kodedninja:
Oh crap, Now bills go up quite easily and unnoticeably :/ Good idea to move the code style to the bottom!

neauoire:
Good idea!

PR #336, opened 2019-10-21
add eti.tf to the list

eti0:
hia, my webring icon is in https://eti.tf/more.html. i hope that's not too much of an issue.

### DIFF
--- a/scripts/sites.js
+++ b/scripts/sites.js
@@ -774,14 +774,6 @@ const sites = [
     type: 'hybrid',
     url: 'https://www.johannesg.com'
   },
-  
-  {
-    author: 'eti',
-    contact: 'eti@eti.tf',
-    feed: 'https://eti.tf/twtxt.txt',
-    langs: ['en'],
-    url: 'https://eti.tf'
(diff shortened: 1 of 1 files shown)


scripts/sites.js
```diff
@@ -773,6 +773,14 @@ const sites = [
...
     type: 'hybrid',
     url: 'https://www.johannesg.com'
+  },
+  
+  {
+    author: 'eti',
+    contact: 'eti@eti.tf',
+    langs: ['en'],
+    url: 'https://eti.tf',
+    feed: 'https://eti.tf/twtxt.txt'
```
ckipp01: Can you make sure your keys are alphabetically in order.

eti0:
is that correct? still learning how to work properly with git.

ckipp01:
@eti0 actually one more comment. It looks like there is a cors issue when attempting to fetch your twtxt file. You'll need to allow the webring to fetch that resource. You can find some info about how to do that in the readme, but let me know if you have any issues getting it to work.

PR #338, opened 2019-10-23
change the location of my twtxt.txt

eti0:


ckipp01:
Looks good @eti0 but before I merge, do you mean to have that random string as an entry?

eti0:
i do!

hint: *it's not random*

ckipp01:
heh, just decoded it, and don't worry, I'm about to be headed to bed. 😏

PR #341, opened 2019-11-13
https://amil.li

nchudleigh:
Webring is in the top right, see image below.  

image

amil.li is a simplification of currentmillis.com

Goal was to make the site very fast and supply the timestamp (in any format) via the path.

I also want people to be able to learn from the code, so I have not minified it and have written it as a single html file.

This is not a personal portfolio or diary or anything.  But it is art.  The simplicity in design and reach for others to learn from the code I think is something that is missing from the web currently.  If this does not fit on webring, sorry- I did read the readme and decided to roll the dice.

### DIFF
--- a/scripts/sites.js
+++ b/scripts/sites.js
@@ -781,6 +781,14 @@ const sites = [
     feed: 'https://y.eti.tf/twtxt.txt',
     langs: ['en'],
     url: 'https://eti.tf'
+  },
+  {
+    author: 'nchudleigh',
+    contact: '@officialneilc',
+    langs: ['en'],
+    title: 'amilli",
+    type: 'tool',
+    url: 'https://amil.li'
   }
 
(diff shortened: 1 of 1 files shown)


ckipp01:
Hey @nchudleigh! Thanks for your interest in the webring!

So this project is really rad, and I'd say that many of us in the ring echo the sentiment of simplification as a form of art. However, we try to keep the webring as a place that can showcase collections of such tools, writings, or art. Ideally, this project would be part of or linked from a portfolio or wiki showcasing it. We try to stray away from individual projects in the ring.

With that being said, if you do have another site which this may be linked from, or decide to start a wiki, portfolio, etc, please let us know and send another pr.

Cheers

nchudleigh:
Fair enough! Figured this might be the case.  Thanks.

PR #346, opened 2019-11-18
added favicon

dym-sh:
[`favicon.ico`](https://raw.githubusercontent.com/dym-sh/Webring/master/favicon.ico) is acly 100x100 .png:

(renamed for backwards compatibility reasons)


neauoire:
Could you make it all black instead of grey?

dym-sh:
problem with just black or just white favicons is they barely if at all visible in several circumstances, like the bookmarks manager or dark-themed OS's

i'll make it black with white outline to account for both

neauoire:
You could put it over a white fill

dym-sh:
just white square looks less stylish

PR #353, opened 2020-01-08
Additional sanitization on channel and tag to prevent onclick injection

virtualsnow:
In the `aside` section, user supplied `val` is used directly in the `onclick` event attribute for channels, users and tags:

Although feed content is cleaned in `escapeHtml`, the page can still be manipulated by bad input since it's already inside `onclick`.

**Example**
`2020-01-20T20:20:20+02:00   /");alert("vs` as an entry in a twtxt file will display an message when the channel is clicked since:
`onclick='filter("${val}")'` is expanded to:
`onclick="filter(&quot;&quot;);alert(&quot;vs&quot;)"` and executes despite sanitization.

I fixed this by whitelisting only alphanumeric values separated by dashes for the channels and tags, in `cleanTags`.  This also limits changes to the message content, where people will legitimately use many weird characters, so the tags are sanitized but the content remains as the author intended.

Since there is no private data or accounts on hallway, this isn't much of a security issue.  However, it should be addressed since it is possible to deface the hallway or accidentally break functionality (unescaped comments, etc.)

### DIFF
--- a/scripts/hallway.js
+++ b/scripts/hallway.js
@@ -171,8 +171,8 @@ function Hallway (sites) {
       const parts = line.replace('  ', '\t').split('\t')
       const date = parts[0].trim()
       const body = escapeHtml(parts[1].trim()).trim()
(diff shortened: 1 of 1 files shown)


neauoire:
Good idea, thanks :)

PR #367, opened 2020-03-10
Add ashpex.neocities.org to webring

Ashpex:
The webring icon is in my footer, at the end of the site.


neauoire:
Welcome to the webring :)

PR #372, opened 2020-04-06
add zvava.org

mininmobile:
👋 

<s>i was planning to do this 3 days ago but then hyperjam happened</s>

the webring icon is at the top left of the homepage


neauoire:
welcome aboard :)

PR #373, opened 2020-04-10
added intimate visual art piece

greg-wolff:
http://www.greg-wolff.com/reach

personal intimate piece regarding death of a loved one, interactive webpage to explore, interim to another webring, the experience will get more interactive and longer over time


neauoire:
I don't think it's a very good candidate for the webring, sorry.

PR #378, opened 2020-05-07
Update sites.js

poryhedron:
Added My link! Best Wishes

https://www.b-liew.com/


neauoire:
I couldn't find the location of the webring icon, also the website doesn't work for me I think, it's just a big background image.

Screenshot from 2020-05-10 08-02-35@2x

PR #391, opened 2020-05-24
Update index.html

florianporada:
Works now also without javascript 😬(~ redirects to a slim version if it detects that js is disabled)
Not sure about the whole single-page thing since a lot of portfolios are kind of single page approaches. 
What is your opinion about it?

...Well I'll give it another shot. ✌️


neauoire:
Hi Florian, 

Sorry but this counts as a single pager as none of the content is hosted on the site itself and only redirects outside, sorry. 

Nice work making it work without javascript tho :)

PR #392, opened 2020-05-24
Add marginchronicles.com

marginalian:
The webring icon is at the bottom of the root page.

### DIFF
--- a/index.html
+++ b/index.html
@@ -412,6 +412,9 @@
       <li data-lang="en nl" id="119">
         <a href="https://proycon.anaproy.nl">proycon</a>
       </li>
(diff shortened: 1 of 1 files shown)


neauoire:
Very nice! It's fun to see that lisp templating in the wild(Emender).

marginalian:
Thanks for the merge! The earlier version of oscean inspired that (and a lot more about my site). But I think Emender's days are numbered now that I'm using Ruby because it has embedded code templating in the standard library.

neauoire:
I have also replaced that system for [something simpler](https://wiki.xxiivv.com/site/meta.html) since.

PR #399, opened 2020-05-28
Hello. Link to webring is in bottom of my site.

llbxg:
ex) https://www.kosh.dev/#webring


neauoire:
I can't find the content, it seems like there is only a page or two.

PR #421, opened 2020-06-25
Added site 128/jlj (provisionally)

jlj77:
I'm displaying the (white) webring icon on my home page: http://ctrl-c.club/~jlj/index.html. (Seems to be loading fine, but I'm happy to host it myself, if you'd prefer.) FYI, I also cleaned up a few of the trailing slashes in the other entries as part of this change; hope that's OK.


neauoire:
Not enough content, missing its own domain.

PR #423, opened 2020-07-01
Add jaks.fi

sjaks:
Add page jaks.fi to the webring. All pages on my site have the web ring icon at the bottom.


sjaks:
Site needs to be looked into and changed a bit.

PR #377, opened 2020-05-06
Added My Link

UserUnknownX:
Thanks For making this Great website

[My link](https://userunknownx.github.io/)

PR #425, opened 2020-07-07
Update index.html

UserUnknownX:
[I'm here again](https://github.com/XXIIVV/webring/pull/377) with a not good-looking [website.](https://userunknown.now.sh)

Maybe It's lack of content, But soon there will be more things.

~~In fact, I'm much more faster than what I thought XD(Maybe just because it's too simple and crude)~~


neauoire:
Nothing loads for me, just a white page saying to use the console?

PR #430, opened 2020-07-24
adding thatshubham to webring

thatshubham:
Will be really glad to join the webring! Nice initiative. The webring icon is added to footer of all pages.

image


neauoire:
Not enough content, everything seems to be linking offsite.

PR #438, opened 2020-08-06
Change PR invite link target.

tyleretters:
Old href pops an edit window. New href links to instructions.


neauoire:
Good idea, maybe people will read the rules more.

PR #444, opened 2020-08-15
added jacob.chvatal.com

jakechv:
The location of the webring icon is in the bottom left of the text content. 
It's stored in the footer at the leftmost position of the main content of the site,
and as such is visible slightly southwest from the center of the web page.

I have self-hosted the icon.


neauoire:
Hi Jake, the website looks nice but it's too new, it's lacking some content(only 2 blog posts, and projects are all documented offsite).

Feel free to apply again when it's a bit further down the line, otherwise it's very nice <3 I'm looking forward to see what comes next.

PR #452, opened 2020-09-06
amorris -> chance

AAorris:
I'm slowly adopting a pseudonym and would appreciate updating my listed website! (twtxt staying the same for now)


neauoire:
I merged this PR because you are already part of the webring, but as your site it right now it fails the requirements to be part of the webring.

AAorris:
Good callout, you're right. I should remove myself for now.

PR #453, opened 2020-09-06
Remove Chance

AAorris:
Not passing the guidelines for the webring right now. Follows #452


AAorris:
Actually, I've updated the page to not use javascript, add writing, and show the side of what's there as the art I see it for :) More to come!

PR #454, opened 2020-09-08
Added deianeira.co to webring, conformed whitespace for preceeding en…

de1ane1ra:
…try.

https://deianeira.co/ webring is placed bottom left of the page in footer.
Also https://deianeira.co/elsewhere/ dead center.


neauoire:
:ok_hand:

PR #460, opened 2020-09-17
Update index.html

mattschrage:
Add schra.ge to webring


neauoire:
I'm struggling finding my way around your website, where is the work you do? about page? I can see a list of bookmarks and things, but I can't find my way into the website.

souvik-03-07-2001:
Can I take the projects? I have done a lot of these stuffs

PR #462, opened 2020-09-19
Update index.html

shouhu21:
add entry for mbaazi.net to index.html


neauoire:
I had a look at the website, it doesn't look handcrafted but instead more like a wordpress theme, the navigation requires javascript, it's missing an about page, etc..

I don't think it's a good candidate for the webring, sorry.

PR #473, opened 2020-10-27
Added odd.codes to webring

iiviigames:
Icon is at the top of the page, right next to the site title.

### DIFF
--- a/index.html
+++ b/index.html
@@ -482,6 +482,10 @@
         <a href="https://metasyn.pw">metasyn</a>
         <a href="https://metasyn.pw/rss.xml" class="rss">rss</a>
       </li>
+      <li data-lang="en" id="137">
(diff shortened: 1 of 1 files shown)


neauoire:
- Single page websites, websites acting only as portals to other social platforms.[failed]
- If your website requires Javascript to display the majority of its content or to navigate, it will be rejected. [passed]
- Your website must count at least 10 pages and include and about page. [failed]
- You must have your own domain name, we do not accept github.io subdomains.[passed]

iiviigames:
@neauoire You're going to have those socks blown off this year. I've spent a year making a masterpiece of that site. I can't wait to resubmit. I feel like such an imposter amongst you all, and for once, I don't. 🎃 is a coming soon 👻 , be ready....

neauoire:
MY BODY IS READY

PR #474, opened 2020-10-29
Add Rishi Kothari (rishi.cx)

rishiosaur:
You can find my website at [rishi.cx](https://rishi.cx); the symbol is just before my name on the first line:

Screen Shot 2020-10-29 at 7 22 29 AM

### DIFF
--- a/index.html
+++ b/index.html
@@ -482,6 +482,9 @@
         <a href="https://metasyn.pw">metasyn</a>
         <a href="https://metasyn.pw/rss.xml" class="rss">rss</a>
       </li>
+      <li data-lang="en" id="137">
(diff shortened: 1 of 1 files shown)


aadilayub:
If you take a look at the [webring criteria](https://github.com/XXIIVV/webring), there's this line:
> If your website requires Javascript to display the majority of its content or to navigate, it will be rejected

With js disabled, I could only see a plain black screen when I tried to load your site. I think you'll have to fix that for your site to join the webring.

Very cool site otherwise though!

neauoire:
Only looks like a blank black screen for me, sorry.

PR #475, opened 2020-10-29
add stefan.midjich.name

stemid:
[My personal homepage](https://stefan.midjich.name) full of mostly technical docs, some recipes and some blog posts. Mixed Swedish and English text.


neauoire:
The webring is centered around hand-crafted websites, despite the great content(esp recipes) your website doesn't qualify as it uses mkdocs defaults and template.

PR #480, opened 2020-11-03
Add notebook.wesleyac.com

WesleyAC:
the link is in the bottom right corner of the main page

there's no about page on my site, but i'd like think that my writing is sufficient to introduce myself :)


neauoire:
I think the introduction post at the end is good enough :) 
Welcome to the webring.

PR #482, opened 2020-11-05
Update index.html

turelio:
added my domain alias, site now actually fulfills the criteria (or will soon as it grows)


neauoire:
Can you also change the url?

neauoire:
perfect :+1:

PR #489, opened 2020-11-28
Add void.cc to the webring

automata:
Hey! I love your projects and the whole rabbits universe.

This idea of a webring triggers good memories from late nights from the 90's, where we had to wait for midnight to connect through a 56k modem and navigate in a world of ASCII zines, IRC channels and loopy GIFs.

I try to keep [void.cc](https://void.cc) as my personal corner in the web with random bits excavated from my brain dump. I humbly hope it fits in yours universe.


neauoire:
Looks like a perfect fit to me :) Welcome to the ring

PR #490, opened 2020-12-02
Update index.html

awavering:
Adding albertwavering.com; linked back from footer.


neauoire:
image

Website renders really poorly for me, images don't load. Any idea on this?

awavering:
Thanks for the check @neauoire - I assumed my static site generator would work fine without js, but that doesn't seem to be true. I'll fix and resubmit!

PR #494, opened 2020-12-12
Update index.html

alessbell:
Add aless.co to webring - link on https://aless.co


neauoire:
It appears that everyone who goes through the recursecenter is incredibly talented.

Btw, there's a blank `<a id='title'>` tag at the top that's invisible, I'm guessing maybe it needs js to display? It took me a while to figure out how to return to the home when there wasn't anything showing up above the post titles. Maybe use an image fallback.

Anyhow, welcome to the webring :)

PR #506, opened 2021-01-11
adding everest-pipkin.com to the webring

everestpipkin:
adding everest-pipkin.com to the webring. my site uses frames (yes, i know, really) so while the webring link is accessible on every page, the icon is actually at https://everest-pipkin.com/related.html


neauoire:
Ah! Amazing! welcome to the webring
I love this itteration of your website(I've been on it many times already)

PR #508, opened 2021-01-21
Added cm3z4.dev to webring.

cm3z4:
I placed the webring logo in the footer. All future pages/posts will be the same. Thank you.

### DIFF
--- a/index.html
+++ b/index.html
@@ -537,6 +537,9 @@
       <li data-lang="en" id="151">
         <a href="https://everest-pipkin.com/">everest pipkin</a>
       </li>
+      <li data-lang="en" id="152">
+        <a href="https://cm3z4.dev">cm3z4.dev</a>
+      </li>
     </ol>
     <footer>
       <p class="readme">


neauoire:
Incomplete.

> If your website requires Javascript to display the majority of its content or to navigate, it will be rejected. Your website must count at least 10 pages and include an about page. You must have your own domain name, we do not accept github.io subdomains.

cm3z4:
Hi @neauoire. Could you elaborate a little more on what I need correct to meet the criteria?

1. My site does/will use JavaScript/JSON to render links for blog/photography posts. I can create a fallback to this method with noscript tags.
2. Working on the content to get the minimum 10 pages. I just launched my site.
3. Do I need a separate about page or is my home page sufficient? I'm building my site a little different than the norm.
4. I do have my own domain name, but the site is hosted on GitHub.

Any feedback helps. Thanks in advance. Looking forward to joining the webring. I'm a big fan of a Hundred Rabbits and Pino.

neauoire:
Sure,

> My site does/will use JavaScript/JSON to render links for blog/photography posts. I can create a fallback to this method with noscript tags.

I cannot see them at the moment, as my browser does not support javascript at this time. It would be good to have them in a noscript too.

> Working on the content to get the minimum 10 pages. I just launched my site.

Don't apply for the webring until your site is complete, wait for a few months before applying so you have interesting content.

> Do I need a separate about page or is my home page sufficient? I'm building my site a little different than the norm.

It has to be accessible by its own url, it can't be a blurb in the footer or something like that.

> I do have my own domain name, but the site is hosted on GitHub.

That's fine :)

Lemme know when you're further down the line with your website so I can have another look!

PR #511, opened 2021-01-23
Add darch.dk to webring - take 2

sorenpeter:
And the icon.white.svg is to be found in the footer.
I hope this is sufficient :)


neauoire:
Looks great! welcome to the ring :) You have a blog post with no date that shows up in 70s on my RSS btw ;)

sorenpeter:
Thanks:) What do you mean with 70s in relation to the blank rss post?

PR #512, opened 2021-01-24
Add natehn.com to webring.

natehn:
Resubmitting! Webring link and icon are in the lower right-hand corner.


neauoire:
welcome to the ring :)

PR #515, opened 2021-02-01
Reduce svg filesize

Seirdy:
This commit reduces the filesize of the SVG files by over half.

I and other siteowners would like to inline the image into our sites;
minifying and shrinking the filesize improves this.


neauoire:
Very very cool, thank you.

PR #518, opened 2021-02-04
added pbatch to webring

PaulBatchelor:
webring logo can be found at the bottom of my homepage here: https://pbat.ch/

I used an id of "pbatch" in the HTML tag, with links to my homepage and twtxt feed.

Thanks!


neauoire:
Welcome to the ring! That'll surely make the Lieu searchable words double haha.

PR #523, opened 2021-02-17
added nnix.com to webring

davidemerson:
I'm only adding the https proxy, though the site is gemini-native.

### DIFF
--- a/index.html
+++ b/index.html
@@ -559,6 +559,9 @@
       <li data-lang="en" id="157">
         <a href="https://njoqi.me">njoqi</a>
       </li>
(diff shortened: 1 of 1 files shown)


neauoire:
Not only is it missing a dedicated about page as required in the rules, I think it's missing content, it doesn't seem to show much creative works, or at least I could not find them. I don't think the site in its current state qualifies for the webring, sorry.

davidemerson:
No worries, I haven't uploaded the portfolio section yet, will get to that and resubmit. Thanks for the check.

PR #525, opened 2021-02-21
Update index.html

rahak:
On the index.html the webring icon is floating on the top right corner.
Rest of the pages it stays in the footer. I completed my site a few days back.
I don't know if I can join the ring, I use bludit for the blog part (sketchbook). 

May the wind be always at your back.
Thanks.


neauoire:
Nice art :)

The website is too early in its development, actual content is mostly sending to instagram and I couldn't locate the about page. Logo doesn't send back to the home, lot of pictures send to work.html instead of project pages.

Please apply again when it's a bit further down the line.

PR #530, opened 2021-03-04
Updating index.html - fixing gr0k entry

gr0k-net:
Apologies, I'm a dunce, and hadn't changed the anchor text :( Fixed now.


neauoire:
I totally didn't spot that one either XD

PR #545, opened 2021-03-10
Add WellObserve (Chinese font fixed)

ChengduLittleA:
I have linked Chinese font needed for the site, so the content should be shown properly. May take few more to load for the first time.

http://www.wellobserve.com/

I'd like to know if it's preferable if I use js to determine browser language and provide a google translate link on my pages?

Thanks!

### DIFF
--- a/index.html
+++ b/index.html
@@ -567,6 +567,9 @@
         <a href="https://www.gr0k.net/twtxt.txt" class="twtxt">twtxt</a>
         <a href="https://www.gr0k.net/blog/feed.xml" class="rss">rss</a>
       </li>
+      <li data-lang="zh-Hans-CN" id="WellObserve">
+        <a href="http://www.wellobserve.com">WellObserve</a>
+      </li>
     </ol>
     <footer>
       <p class="readme">
--- a/index.html
+++ b/index.html
@@ -578,6 +578,7 @@
         <a href="https://www.gr0k.net">gr0k</a>
         <a href="https://www.gr0k.net/twtxt.txt" class="twtxt">twtxt</a>
         <a href="https://www.gr0k.net/blog/feed.xml" class="rss">rss</a>
+      </li>
       <li data-lang="en" id="clarity">
         <a href="https://clarity.flowers">Clarity Flowers</a>
       </li>


ChengduLittleA:
The page has been updated for translation options and a better gallery page

neauoire:
Ah the gallery icon helps a lot :) I still can't see the font tho, where is the translation option?

image

Also, would you like to add yourself a 88x31 button?

neauoire:
I see a `transbtn` element in the <li> but it's empty inside.

ChengduLittleA:
Thanks for the feedback. It seems the font might be too big to load (an avera chinese font file is like 20MB). I'll incorporate a font reduction method to ensure it loads relatively quickly. What system and browser are you using? I'm not sure why the page looks like that, maybe on Ubuntu I have multiple language packs already installed. Is it possible you could pull out the data traffic page from F12 in your browser to see if anything is wrong with that font data? Thanks!

I used js to determine client browser's language, since js should not be required to view the content, I'll change it to a static "translate to English" button and if js is running, I'll replace with desired language. It redirects to Google translate though, so I imagine there will be js involved anyway.

neauoire:
I mostly use netsurf, but since your website didn't look too good on it, I installed firefox. The pictures from above are from firefox, I don't have chinese fonts installed, I have to look into that.

It's possible that my internet connection blocks the font, we use a satellite phone that's pretty restrictive in what it can download.

On netsurf, your site looks like:

2021-03-16-084554_1366x768_scrot

ChengduLittleA:
Oh if you are using satellite phone, please don't bother wait for the font to finish loading... it's 23 MB in size and browser sometimes doesn't cache all that. Sorry for the trouble, Chinese has seen trouble since the dawn of computer age and the font indeed doesn't really work that well even for the sites we browse day to day🤔.

On the other note though... it might be a good practice to install "some" basic language packs so right now I'm able to view Russian or Japanese stuff "reasonably okay", and this saves the overall transmission for the internet. It's probably not a very good idea especially for low power use cases to handle large resources, but since I believe you don't really view Chinese content that much, it doesn't really matter to pass on that XD. Id suggest installing some language support for the OS when you guys have Wi-Fi connection :D but for the moment I guess I'll clean up the gallery so that it doesnt need js to open up the whole image.🤗

PR #564, opened 2021-03-15
Added my personal blog, cannedfi.sh

chilledtonic:
This webring inspired me to finish and relaunch my personal blog, and I am excited to be listed!

You can find the webring link-back on the footer of every page in the website; next to the other "Social" icons.

Let me know if I need to edit anything else / did anything incorrectly.

### DIFF
--- a/index.html
+++ b/index.html
@@ -591,7 +591,7 @@
         <a href="https://eric.jetzt/twtxt.txt" class="twtxt">twtxt</a>
       </li>
       <li data-lang="en" id="159">
-        <a href="https://cannedfi.sh/">canned fish</a>
+        <a href="https://cannedfi.sh">canned fish</a>
         <a href="https://cannedfi.sh/posts/index.xml" class="rss">rss</a>
 </li>
       <li data-lang="en" id="gr0k">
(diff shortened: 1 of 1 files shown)


neauoire:
> If your website requires Javascript to display the majority of its content or to navigate, it will be rejected. 

2021-03-16-084257_1366x768_scrot

It's a bit odd, your `<header>` has a whole bunch of stuff in it, but it doesn't show up at all? 

> Your website must count at least 10 pages and include an about page.

I seems your website only has 3 blog posts and no pages, the about page is inaccessible without javascript(?!)

chilledtonic:
This was very strange to me, since the website only makes a handful of Javascript calls, none of which actually affect navigation in any way - just some minor page features; namely, showing what I last listened to from ListenBrainz.

Then I realized that the icon library I used insert themselves with Javascript, which is why the navigation header is missing. I'll replace this library and submit again.

Any advice for a noscript browser so I can recreate your browsing environment?

Thanks.

neauoire:
I recommend Netsurf :) It's an excellent browser.

PR #568, opened 2021-03-17
Add aj booker to webring

aj-booker:
webring icon can be found in the footer of every page

https://ajbooker.co

<33


neauoire:
> Single page websites, websites acting only as portals to other social platforms, or websites with violent, racist, sexist or speciesist content will be rejected.

Failed.

> If your website requires Javascript to display the majority of its content or to navigate, it will be rejected. Your website must count at least 10 pages and include an about page. You must have your own domain name, we do not accept github.io subdomains.

Failed.

PR #572, opened 2021-03-28
Add sirtetris.com

IllDepence:
Added [sirtetris.com](http://sirtetris.com/).

The webring icon/link is at the bottom of all content pages (not the index page).

There's a small about section [here](http://sirtetris.com/?c=misc) and the [hobby section](http://sirtetris.com/?c=hbby) is also in part kind of a self introduction.

### DIFF
--- a/index.html
+++ b/index.html
@@ -607,6 +607,9 @@
         <a href="https://マリウス.com">マリウス</a>
         <a href="https://マリウス.com/index.xml" class="rss">rss</a>
       </li>
+      <li data-lang="en de" id="163">
+        <a href="https://sirtetris.com">sirtetris</a>
+      </li>
     </ol>
     <footer>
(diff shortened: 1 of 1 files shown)


neauoire:
Very nice website :) Just a comment, I love the content, but I hate how these 4 pages are totally disconnected from each other. It would be much more fun to be able to not have to go back to the empty splashpage to discover the other related interests.

IllDepence:
> It would be much more fun to be able to not have to go back to the empty splashpage to discover the other related interests.

Totally agree. Having a proper navigation on the page fell victim to me putting form over function by introducing the logo/slogan/splashpage at some point.

Looking through the webring today I learned about now pages. Maybe I'll figure something out to reintroduce a navigation and then be able to also add a now page. (:

PR #573, opened 2021-03-28
Add rodrigofranco.com

caffo:
Added webring link to the footer.

@neauoire I imported content from many of my other sites/sources and organized the homepage to make it clear. Sorry about being a pain, but I really think my site deserves to be part of the webring — I feel like I found my tribe there :)

Can I get reconsidered? 🙇

### DIFF
--- a/index.html
+++ b/index.html
@@ -610,6 +610,9 @@
       <li data-lang="en de" id="163">
         <a href="https://sirtetris.com">sirtetris</a>
       </li>
+      <li data-lang="en" id="rodrigo">
+        <a href="https://rodrigofranco.com">rodrigofranco</a>
+      </li>
     </ol>
     <footer>
       <p class="readme">


neauoire:
Sure :) I'll be back on the boat soon, I'll have another look :)

caffo:
Thanks! I'm devouring 'Busy Doing Nothing' this weekend BTW.

neauoire:
I've had another look and it still doesn't qualify since there is so few projects, but if you want I can give you some pointers, since it's almost there. Would you be open to some feedback for your site, we can keep this PR open while we do this until it's ready to be merged. :)

Lemme know if you like the idea and I'll give you some ideas.

caffo:
Absolutely. I would be delighted.

neauoire:
Alright! I'm glad :)

So, I understand you like the Less Is More concept, your website is quite minimal, but its lack of form, it doesn't make its function justice. A problem I see in a lot of website on the webring is that everyone is going for ultra minimal websites, which in the end all looks the same, plain text on plain background, which is not a problem in itself, but it does not invite the visitors into its world at all. And the webring, ideally, is like portal into people's worlds, so some personality has to emerge, even if the website wants to be minimal.

Your website has very poor navigation, and the categories are all over the place. So here's what I suggest:

- You have 2 "content" pages(you need 10 for the webring btw), whois(I recommend renaming to /about), and /now.
- Projects: You only really have 1 project, and it's not public, and not documented, the game engine. You have one game for it and it's not documented on your website it sends to a 3rd party. So, remove projects until you have some.

- About: You have clickable words for almost everything in this page, thinkpad, music, Brazil, tangible, should all be clickable and ways to enter the website.
- Now: Your missing at least a timestamp to show it's an active page, see [/now](https://ftrv.se/now) for inspiration. Almost everything in your now page doesn't go there.
- Focus on notes since you're clearly using this more like a notebook. Use your header to go backward to the parent topic, or get rid of it. I suggest your either give it a purpose, like a breadcrumb navigation, or at least a way to go to the principal portals.
- Your website is super linear and there's no way to go back, and pages aren't interlinked. So I recommend you make a sort of orphan/deadend finder thingy, that tells you which page contain less than 2 links to other pages and flag them as deadends, and which pages are only linked to by less than 2 pages and flag those as orphans. Then make sure you have neither orphans and deadends.

caffo:
Thanks for your comments. One in special puzzles me a bit, regarding having only 2 content pages. What defines a content page in your view? Isn't the [Resonance Log](https://www.rodrigofranco.com/Resonance_Log.html) a content page? 

Also my definition of a project is anything I'm willing to pursue and keep updated.  In that case the list of [Meditative Games](https://www.rodrigofranco.com/Meditative_Games.html) is a project, and my [Trips](https://www.rodrigofranco.com/Trip_Notes.html) and [Yearly Notes](https://www.rodrigofranco.com/Yearly_Notes.html) as well.

I agree on the other points. WIll work on them. Thanks again :)

caffo:
Made some initial tweaks:

* Added the updated timestamp to the `Now`
* Removed `Whois` all together by merging it with the home page, giving it a more descriptive format. 
* Made the "Back to home page" link in the header more clear. The old header linked to root, but was hard to notice.
* Removed the game engine for now.
* Sprinkled links around the `Now` page.

Plan to tackle on the upcoming days/weeks:

* Add a page for my software company, replacing the external link I currently have on the home page.
* Link pages where possible. 
* Deal with orphans and dead ends.
* Implement a dynamic breadcrumb system that works with the notebook/garden-y flexibility of the website.

Let me know what I did until now is enough to be added or if you want to wait until the other changes are implemented to accept the PR.  Feel free to comment more if you feel it would make it better. 

Again, I'm very grateful for the advice you provided, even if I never make into the webring :)

neauoire:
You know what, your website is well on it's way, welcome to the webring :)
Thanks for taking the time to make the changes <3

neauoire:
I'd suggest you add now and about to your header until you complete the breadcrumb navigation.

caffo:
Thanks yay! Will do it.

PR #575, opened 2021-04-04
add oppen.digital to webring

fiskurgit:
Webring icon is in the footer: https://oppen.digital


neauoire:
Welcome to the webring :D

PR #577, opened 2021-04-09
Add c-sadovnikov to webring

csadovnikov:
Hi, I would like for my site (c-sadovnikov.com) to be added to the webring.


neauoire:
- Couldn't find webring icon.
- Incomplete PR.

PR #590, opened 2021-05-19
add kevinnel.com

kevin-nel:
Hi, I'd like to have my site added to the webring.


kevin-nel:
I see now from reading some other pull requests that blog posts aren't counted as pages. I don't think my site qualifies then. I will open another pull request once I have added some more independantly standing content on my website.

PR #591, opened 2021-05-20
Added sakrajda.eu to the ring

przsak:
added my website https://sakrajda.eu


neauoire:
Everything seems to be hosted offsite and linking to 3rd parties, I couldn't find a content page.

PR #597, opened 2021-05-25
added tommi.space to the webring

xplosionmind:
I would be honoured to join the XXIIVV Webring.
The link to the webring is located [in the footer of every page](https://tommi.space/home#0 'the bottom of tommi.space homepage')).

Best,
T

### DIFF
--- a/index.html
+++ b/index.html
@@ -625,6 +625,11 @@
       <li data-lang="en" id="nilfm">
         <a href="https://nilfm.cc">nilFM</a>
       </li>
+      <li data-lang="en it" id="tommi">
+        <a href="https://tommi.space" target="_blank" title="A virtual representation of Tommi’s brain">tommi.space</a>
+        <a href="https://tommi.space/feeds/zibaldone.xml" target="_blank" title="Blog posts RSS feed" class="rss">rss</a>
+        <a src="https://tommi.space/logos/favicon.gif" title="Tommi’s head exploding" />
+      </li>
     </ol>
     <footer>
       <p class="readme">


neauoire:
Seems like this requires javascript? The site looks very broken.

2021-05-25-085859_1366x768_scrot

I cannot see the webring icon in the footer either?

2021-05-25-085853_1366x768_scrot

xplosionmind:
I think you are not loading the CSS file correctly.  
The only two things which depend on JS are the menu on mobile and the theme switcher, therefore you should see the website as I do. When I disable JavaScript in my browser, the website does not look broken.

Screenshot 2021-05-25 at 7 04 14 PM

The CSS file, as you can see from [the source code of the page](https://github.com/xplosionmind/tommi.space/blob/8374ee142b77be232edfbdba99ffc1424ebd89c2/_includes/head.html#L16), is normally loaded with an HTML `link` tag. I really do not understand the reason why you are experiencing this trouble by browsing it.

About the webring icon missing, my bad. I probably forgot to build the site.   
Please, [check again](https://tommi.space/home#0 'the bottom of tommi.space’s home')

neauoire:
https://user-images.githubusercontent.com/1455726/119541616-79de4380-bd43-11eb-9682-f6bf886d5397.mp4

Maybe your website uses a lot of CSS3 things? my browser doesn't support any of it. Firefox is very slow on my old computer, I tried loading the page but there was a sort of animation that was slowing down the computer a lot, I had to close it. 

I'm not sure why I can't see any of the colors tho, that should be working.. I'll do some digging.

xplosionmind:
Yep, my CSS is packed of CSS3 stuff. I honestly do not know what is CSS2 and what is CSS3, but I am surprised that without the latter the website looks so bad.

neauoire:
After some digging I realized that the use of variables is why the website looks all broken for me. 
I guess I hadn't considered that before, it's the first time I see a website that relies so heavily on CSS3 that it's unusable for browsers that only supports vanilla CSS. Do you have plans to make your website accessible to non modern web browsers? And remove the javascript banner at the top?

xplosionmind:
> Do you have plans to make your website accessible to non modern web browsers?

Unfortunately, no. I am not a skilled developer and this variables solution is optimal to me. CSS3 is awesome and I need to exploit it at its full capacity.

<br />

> And remove the javascript banner at the top?

I am considering this. I probably will.

Sorry for the late response, I thought a lot about this…

PR #604, opened 2021-06-13
Added embyr.sh

missembyr:
Feel free to reject as my site doesn't meet the 10 content pages criteria yet however I feel like it is in spirit of the webring and will only grow from here.

I have also used italic styling which, while I don't think it was mentioned in the rules I've not seen anyone else use apart from ƒdisk's usage of the ƒ. I'm happy to remove this if required.

As I am aware that you're using a satellite connection I thought I should also mention that the [photography](https://embyr.sh/photography) sub-page "Liverpool" is around 10mb large due to the high resolution images.

Let me know if there are any issues.

### DIFF
--- a/index.html
+++ b/index.html
@@ -628,6 +628,10 @@
       <li data-lang="en" id="pixouls">
         <a href="https://pixouls.xyz">pixouls</a>
       </li>
+      <li data-lang="en" id="embyr">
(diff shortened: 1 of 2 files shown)


neauoire:
It's well on its way to become a full website :) Welcome to the webring!

PR #607, opened 2021-06-15
added tomfasano.xyz

tfasano1:
The webring link/logo is in the bottom nav bar.

### DIFF
--- a/index.html
+++ b/index.html
@@ -635,6 +635,10 @@
       <li data-lang="en" id="maxhaesslein">
         <a href="https://www.maxhaesslein.de">maxhaesslein.de</a>
       </li>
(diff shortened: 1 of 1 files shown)


neauoire:
I have a few issues with your website
1) I can't find your works? I can find 3 tracks, the videos page requires js so I can't visit it now, and comics is a single image.
2) Almost all topics are empty, unclickable or coming soon.
3) In your links you point to Luke Smith which is a massive imbecile, and I would be afraid that someone accidentally ends up on their web page coming from the webring. 

I'll have a to decline your application, I don't think it's a good fit for the webring in its current state sorry.

tfasano1:
Ok, I'll start my own.

PR #611, opened 2021-07-02
Update index.html

paulbricman:
The webring icon is located in the top-right corner


neauoire:
Sorry, your website is not a good candidate for the webring at this time.
It does not have a dedicated about page, and does not have at least 10 projects documented on site.

paulbricman:
Thanks for feedback, I'll get back when the requirements will be fulfilled!

PR #615, opened 2021-07-13
Added jonathanxu.com

JonathanXu1:
Thanks for weaving the ring! Adding a tiny thread to the web.
The icon is located on the left hand menu

### DIFF
--- a/index.html
+++ b/index.html
@@ -642,6 +642,9 @@
         <a href="https://compudanzas.net/tw.txt" class="twtxt">twtxt</a>
         <a href="https://compudanzas.net/atom.xml" class="rss">rss</a>
       </li>
+      <li data-lang="en" id="jonathanxu">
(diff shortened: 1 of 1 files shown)


neauoire:
Hi Jonathan,

Thank you for your application, but your website is not a good candidate for the webring in its current state. None of the content is clickable, looking at the source it seems they are div? and not links. Furthermore, there is no about page, and the cooking page has pictures of dead animals.

JonathanXu1:
Thanks for letting me know, I'll make the divs into links and add a tab to the about page, which is the home page. However I'd like to say that the requirement about the cooking page seems arbitrary in that it wasn't specified anywhere on the Readme so I wasn't aware.

neauoire:
Maybe a content warning, maybe a photo that is some sort of substitute, and in the recipe itself it could say that it works equally well with chicken, something like that. :) Let me know when you've pushed the changes

PR #620, opened 2021-07-26
add http://alphamethyl.barr0w.net/~skye/

monarrk:
hi! I am a big fan of the sites on this webring so I thought I'd try to join it

the webring is on http://alphamethyl.barr0w.net/~skye/README.html

I hope that it has enough content to be considered 10 pages worth. I'm not sure if a bunch of scattered picture, video, and document files can be each considered a page on its own, but maybe it could be

### DIFF
--- a/index.html
+++ b/index.html
@@ -654,6 +654,9 @@
         <a href="https://compudanzas.net/tw.txt" class="twtxt">twtxt</a>
         <a href="https://compudanzas.net/atom.xml" class="rss">rss</a>
       </li>
+      <li data-lang="en" id="skye">
+        <a href="http://alphamethyl.barr0w.net/~skye">alphamethyl.barr0w.net/~skye/</a>
+      </li>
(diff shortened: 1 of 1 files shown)


neauoire:
Hi Monark, 

Sorry but your projects are all documented offsite(soundcloud, bandcamp, etc..), this is not a good candidate for the webring.

monarrk:
hm, well, I'm not sure if you missed it because it's a bit not very easy to see, but if you hit the little "../" link at the top it brings you to the actual main page http://alphamethyl.barr0w.net/~skye/ , which contains on-site original work which I have made. a lot of it is mixed in with not-original work though with no clear indicator so I am sorry about that. if you did already see that, though, then I accept my fate

neauoire:
Yeah it's a bit of a mess right now, but if you turn this into a personal website in the future, I'd love to consider it again. :)

PR #621, opened 2021-08-02
Update index.html

sadness97:
Added https://letscrushcapitalism.neocities.org to the webring. Thanks!


neauoire:
Interesting content, but not a good candidate for this webring as it focuses on personal websites.
Missing an about page, also lots of [broken links](https://letscrushcapitalism.neocities.org/garden.html#al).

PR #646, opened 2021-12-02
Add jzhao.xyz

jackyzha0:
Webring logo is bottom left on the website footer :'))

Such a wonderful group of websites, I always come back to this webring for inspo

### DIFF
--- a/index.html
+++ b/index.html
@@ -673,6 +673,9 @@
         <a href="https://www.paritybit.ca">paritybit.ca</a>
         <a href="https://www.paritybit.ca/feed.xml" class="rss">rss</a>
       </li>
+      <li data-lang="en" id="jzhao">
+        <a href="https://jzhao.xyz">jzhao.xyz</a>
+      </li>
     </ol>
     <footer>
       <p class="readme">


neauoire:
> Single page website, all projects are documented offsite.

jackyzha0:
did you check https://jzhao.xyz/toc/directory/

neauoire:
This should be the link that is used for the webring, it's a lot more in line with what the webring is for.
But even knowing about this page, I can't find a single link from the home page that will send me there?

jackyzha0:
image
Each of these link to a page and there is an explicit link to that page in the second paragraph: 
image

neauoire:
2021-12-05-171846_1366x768_scrot

I don't see the nodes at the back, I only see a few anchors and they point to the same page, I'm not sure how to get to what you see?

jackyzha0:
> 2021-12-05-171846_1366x768_scrot
> 
> I don't see the nodes at the back, I only see a few anchors and they point to the same page, I'm not sure how to get to what you see?

Oh that's definitely not what it's supposed to look like haha. Do you have JS disabled?

neauoire:
Yeah, one of the requirements for the webring is that the website be usable without javascript.

jackyzha0:
ah I see. I'll delete the branch then!

PR #647, opened 2021-12-07
Add tham.es

richardthames:
The webring icon is centered at the bottom of every footer (except on the 404 page).


neauoire:
Website is too early in its development, open another PR when there's 10+ pages in the wiki section and most link in the about page don't link offsite.

PR #657, opened 2022-01-07
added polarhive.ml

polarhive:
Added my website & it's RSS feed to the webring. 
The link to the icon is at the ``<footer>`` of every page. Let me know if I missed any page.


neauoire:
I'm trying to find your works, but I can't find a single project page on your site, you have a couple of pages on something called codeberg which requires javascript, pictures on pixelfed, which ALSO require javascript, and same for video. The website is lacking documentation that is onsite.

PR #661, opened 2022-01-18
Add duckpuck.net to the webring

girardof:
Submitting my freshly made personal website, it's a mix of portfolio and personal wiki. [The icon is present in the footer](https://duckpuck.net/). 

I like the idea of circular linking as described in the README but there isn't any link after the one I'm adding at the moment therefore I picked one that I find inspiring.

Thank you for the review and for the project.

### DIFF
--- a/index.html
+++ b/index.html
@@ -678,6 +678,10 @@
       <li data-lang="en" id="d6">
         <a href="http://plastic-idolatry.com/erik">Plastic Idolatry</a>
       </li>
+      <li data-lang="en" id="duckpuck">
+        <a href="https://duckpuck.net">Duckpuck</a>
+        <a href="https://duckpuck.net/atom.xml" class="rss">rss</a>
+      </li>
     </ol>
     <footer>
       <p class="readme">


neauoire:
Hi Florian,

The website is nice, but while there are a lot of backend and frameworks projects, I can't find any of your artistic works, I like to think of the webring as a platce that gathers the cross-over between art and code, but I can't really find much art on your site, you have a few gui apps documented offsite, but not has any graphics, are you planning on sharing some of that aspect of projects as well?

girardof:
Hello Devine, thank you for taking the time to visit my website. 

For now most of my visual art falls under the category of glitchart, I will share more about those in the future for sure. 
However I enjoy creating tools that can be used by others and to see what uses they have of it more than to use them myself. So the website will mostly be about the tools themselves. 

Regarding the GUI you are referring to I'm working on rewriting them as they were created using a GUI that is creating binaries too big for my taste and is now an abandonware therefore useless at the moment.

neauoire:
The site as it is now, focuses a lot on tools and doesn't have any information about the work done with them, I tried looking for the result created from the tools and it was pretty hard to find, or demos were hosted off-site and required javascript to be visible.

I try to keep the webring around personal websites where it's not only about the tools but also the context around them and their creator. Your website in its current state is not a good fit for the webring, but if you do add more context to these nice projects, I'd love to have another look :)

girardof:
Thank you for the sensible explanation, maybe at another time then 😉.

PR #666, opened 2022-02-02
add jakobmaier.at

gue-ni:
The icon is in the footer.


neauoire:
Nice blog, but: 

- I couldn't 10 projects that were documented on-site and didn't require javascript to show.
- The about page is lacking information.

PR #667, opened 2022-02-03
Add detondev.neocites.org to webring

DetonDev:
[https://detondev.neocities.org/landing.html](https://detondev.neocities.org/landing.html)
The webring icon is in the footer of every page.

### DIFF
--- a/index.html
+++ b/index.html
@@ -670,7 +670,7 @@
       </li>
       <li data-lang='en' id='detondev'>
         <a href='https://detondev.neocities.org/landing.html'>DetonDev</a>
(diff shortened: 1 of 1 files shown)


neauoire:
Hi Sam :) It's nice to see a PR from you :wave: 

I don't allow for subdomain in the webring, do you have plans of getting your own domain?

DetonDev:
> I don't allow for subdomain in the webring, do you have plans of getting your own domain?

Hmm...In my household, I can't spend any of my money without permission from my parents until I am an adult, so I can't purchase a domain as they would know about it. Though I do intend to get a domain when I can, that would be at least 1.7 years away :/

PR #675, opened 2022-02-23
added nnix.com (gemini proxy)

davidemerson:
This is my personal site for journal, hobbies, projects, and DJ sets. It's really a proxy of a Gemini capsule, so visit on that protocol for the intended experience!


neauoire:
It's a bit too early in its development, it fails a few requirements for the webring in its current state, sorry.

PR #691, opened 2022-04-20
👻 Added the oddblog

iiviigames:
The webring logo is found at the top left of the page, first in line on the navigation bar.


iiviigames:
I'm moving content from a different part of the site here so if it hasn't reached ten unique pages by tomorrow, just give it another day or two!

neauoire:
Welcome to the webring :D

PR #694, opened 2022-05-01
add Yekai

xiamuguizhi:
add Yekai


neauoire:
Sorry but the website is missing a hand-crafted feel, it seems that it is only a blog using a default template in its current state. It's not a good candidate for the webring.

xiamuguizhi:
@neauoire 
Manau, kad suprantu, ką sakote, ir tikiuosi, kad ateityje bus dar viena galimybė.

PR #696, opened 2022-05-03
Updating mineral existence banner icon

flber:


flber:
wow I'm sorry, usually I don't make such obvious mistakes! :P
guess the sleep deprivation is really kicking in, I'm just going to change the file names on my site...

PR #712, opened 2022-06-25
name update for site

armaina:
Embarrassingly, I forgot to add the correct name
fixed name on line 709


neauoire:
Haha, that's okay, I should have double checked XD

PR #713, opened 2022-06-27
add sindhu.live to webring index

sindhusgithub:
Hi, hoping to have my digital garden added to the webring index!
The webring icon can be found at the bottom of the about page: https://www.sindhu.live/about


neauoire:
Nice digital garden Sindhu, welcome to the webring :)

PR #720, opened 2022-08-02
adding aphrodite.dev + security fix?

Arteneko:
Hi!
The backlink is available on the website's footer, and also on [the link garden](https://www.aphrodite.dev/~resources/) page.

I also included a patch for the README, to avoid the bad behaviour that happens without `rel=noopener` (page takeover). It won't fix if people leave that out on their own website, but the easy copy/paste won't have this issue anymore at least.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@ This [webring](https://wiki.xxiivv.com/webring) is an attempt to inspire artists
 ## Join the webring
 
 ```html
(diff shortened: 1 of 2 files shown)


neauoire:
Thanks for the fix to the link snippet :) 
Nice website! I've visited it a few times in the past, welcome to the webring.

PR #721, opened 2022-08-05
Added c16.one to the index

C161:
I added the icon + the link which both are in the header of [https://c16.one](https://c16.one)


neauoire:
The website is too early in its development, it's missing a few pages to validate for the webring, keep updating it :) feel free to open a new PR when it's further down the line.

PR #723, opened 2022-08-11
Adding Kaeforest for the webring

jxnosila:
I'm about to fill the pages with the content I desire, and will probs do so in a few days, also to include research into my given areas of synthesis.
Though hopefully this will be good for the time being :)
https://kae.si/home


neauoire:
Nice wiki :) welcome to the webring!

PR #734, opened 2022-10-05
Adds kira.solar

hackergrrl:
You can find the webring icon in the footer (right side) of every page! http://kira.solar


hackergrrl:
Thanks!

PR #735, opened 2022-10-06
Update index.html

Jim707t:
the logo is at the footer of the blog


neauoire:
Failing passing these criterias

> You must have your own domain name, we do not accept github.io subdomains.
> If your website requires Javascript/CSS3 to display the majority of its content or to navigate, it will be rejected.

PR #739, opened 2022-10-20
added emile.space

HanEmile:
After rewriting my static site generator [vokobe](https://git.sr.ht/~hanemile/vokobe) in the beginning of the year, [emile.space](https://emile.space) has reached a form in which I'd like to submit it. I've tested it using `netsurf` where it looks quite rudimentary, although still fully functional (I'd argue it looks even cleaner completely without css in `lynx`). 

The webring icon is located at the bottom right of each page in the footer.

### DIFF
--- a/index.html
+++ b/index.html
@@ -752,6 +752,10 @@
 			<li data-lang="en" id="chotrin">
 				<a href="https://chotrin.org">chötrin's wiki.</a>
 			</li>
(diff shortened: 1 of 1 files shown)


neauoire:
Very nice website! Lots of projects to get lost into, welcome to the webring :)

PR #742, opened 2022-11-09
add pfy.ch

pfych:
Programmer & Photographer, I have a `/now.html` and `/twtxt.txt` I'm hoping to begin to be more active on but right now its mostly posts about my Photography. Hoping to write more about the work I'm doing code wise soon.


neauoire:
The website is a bit bare in its current state, I tend not to allow sites that are strictly blogs in the webring, most of the links on there send to 3rd party websites(twitter, etc..) which require javascript, the photos page is MASSIVE, I couldn't load it all and at a glance it didn't seem like there was any description for them, just a big photo dump.

You're welcome to open another PR further down the line when the site has a bit more content.

PR #753, opened 2022-12-01
add bloguslibrus

Bridouz:
I am a mere unix user, loving minimal and well designed softwares.

My website is in french, shared with my dad about writing on music releases we listen to.
It also contains a series of prose about my job (nursing assistant), a series of follow-up short stories written alternatively with my dad and a digital garden.

The latter is in its early stage.

Thanks for considering


neauoire:
Il est cool ton site :) Bienvenu sur le webring!

PR #766, opened 2023-01-22
Add stevegattuso.me

stevenleeg:
Adding [my website](https://www.stevegattuso.me) to the ring, if you'll allow it :). I don't have an explicit about.html but I do include what its contents would be at the top of every page. Happy to add a real page in if that doesn't fly though.

You'll find the webring icon on the footer of all pages.

### DIFF
--- a/index.html
+++ b/index.html
@@ -778,6 +778,10 @@
 				<a href="https://guerrero.ph">guerrero.ph</a>
 				<a href="https://guerrero.ph/rss.xml" class="rss">rss</a>
 			</li>
+			<li data-lang="en" id="stevegattuso">
+				<a href="https://www.stevegattuso.me">stevegattuso.me</a>
+				<a href="https://www.stevegattuso.me/feed.xml" class="rss">rss</a>
+			</li>
 		</ol>
 		<footer>
 			<p class="readme">


neauoire:
Hey steve :) good to see a PR from you.

At the moment, I can count only about 3 pages of content(2 projects+photos). The website is coming along nicely but it's a bit too early in its development to be part of the webring. The reason I ask for an about page is so I have a clear place to go to and contact you, but since you have an email contact in the header, it'll be alright for the about page.

You can keep this PR open until the website has more content, or re-open it when you're ready.
Keep me posted!

stevenleeg:
Oh interesting- any reason why the pages nested under "Lists", "Tech Stack" and "Photos" wouldn't count towards the 10? Each of the nested pages would contribute:

* Photos: 6 galleries
* Lists: 2 pages
* Tech stack: 5 pages (maybe 6, but one if them technically links to a blog post so it'd be out based on the criteria)
* Projects: 2 pages

Maybe important to note that all but the photo galleries aren't blog posts- they're wiki-style pages that I try to keep updated (hence the `updated at` datestamp on the bottom).

I could be totally misunderstanding what counts as a page though. Just want to clarify so I know whether or not it's worth re-opening in the future.

neauoire:
I tend not to count tech stack pages, everyone has those and they're quite uninteresting. The issue I have with those was that I used to have these applications that was almost all tech stack rambles, and pretty much not artistic projects of any kind. I look for personal projects with some form of artistic sensibility documented onside(not on 3rd party github pages).

stevenleeg:
No worries at all and thanks for the explanation. Sounds like my site isn't a good fit just yet so I'll go ahead and close this PR.

PR #768, opened 2023-01-25
Add plungepool wiki

plungepool:
Adding my new wiki for consideration, webring icon is in the footer: https://wiki.plungepool.dev

I've been a reader and fan of the xxiivv wiki for a couple years now and always really wanted to create something like that for myself, hopefully it is acceptable in its current state. I have about 15 content pages so far not including the organizational "hub" pages that are lists of links to other pages. I wrote a simple custom static site generator to manage the site which can be found here - https://github.com/plungepool/wiki-dot-plungepool-dot-dev

### DIFF
--- a/index.html
+++ b/index.html
@@ -779,7 +779,7 @@
 				<a href="https://guerrero.ph/rss.xml" class="rss">rss</a>
 			</li>
 			<li data-lang="en" id="plungepool">
-				<a href="https://wiki.plungepool.dev/">plungepool</a>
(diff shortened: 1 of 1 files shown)


neauoire:
Very nice wiki! Welcome to the webring :)

PS. You should reduce the image sizes, and compress them for web, this was brutal to load.

plungepool:
Thanks!! And thanks for the heads-up, I will work on that asap and maybe find some way to automate that.

PR #773, opened 2023-02-09
Eli oat badge fix

eli-oat:


neauoire:
Can you make a PR that only modifies your entry?

eli-oat:
Lol, I hope so! I thought that is what I was doing, but I guess I done and goofed! I'll give it another try!

PR #781, opened 2023-03-10
Adding tosatur.com

tosatur:
I've redesigned my site, (used to be martinpetkovski.com) It now contains no javascript.

The webring link is on the sidebar.


neauoire:
The site seems very much early in its development, most of the projects are quite empty at the moment or documented offsite. I think you should hold off until there's more content, but otherwise it's coming along really well. Please open a new PR when the photography section has been added and when projects include a bit more details than a paragram and a itchio link

PR #791, opened 2023-03-28
add grgv.xyz (now with webring icon/link), fix indentation

coolvision:
Webring icon/link on grgv.xyz is in the footer, rightmost element.
Sorry did not get this right first time )


neauoire:
- Could not find 10 non-blog pages that are documented on-site, all documented seems to be linking to third parties like github.

PR #798, opened 2023-06-13
Update index.html with a new website

red47:
I added the webring icon on the index file. At the bottom.

### DIFF
--- a/index.html
+++ b/index.html
@@ -767,6 +767,9 @@
 				<a href="https://pfy.ch">pfych</a>
 				<a href="https://pfy.ch/rss.xml" class="rss">rss</a>
 			</li>
+			<li data-lang="en" id="47nil">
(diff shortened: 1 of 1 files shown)


red47:
Wearing image is at the bottom of the index file

neauoire:
Hi red47,

Your website in its current state is not a good candidate for the webring. I try to curate it with websites that have a bit more than a blog(more on the wiki side of things) but right now, the project page has blurbs but no project pages. I require websites to have at least 10ish topic pages that do not include blog posts.

red47:
I understand, thank you.
in fairness, a lot of the sites on the webbing currently only have a page or two, if at all.

neauoire:
I noticed, but during their application, they had each at least 10 topic pages. X(

PR #800, opened 2023-06-16
Remove broken URLs

ubuwaits:
This PR removes the following URLs:

**URLs that return a 404 error:**
- https://sixey.es/feed.xml
- https://0xff.nu/feed.xml
- https://milofultz.com/atom.xml
- https://wilde-at-heart.garden/notes/feed.xml
- https://arcades.agency/links/rss.xml
- https://log.tone.support/log.xml

**URL with certificate issue:**
- https://canalswans.commoninternet.net/rss.xml

**URLs where the server is no longer responding:**
- https://guerrero.ph/rss.xml
- https://guerrero.ph/

### DIFF
--- a/index.html
+++ b/index.html
@@ -289,7 +289,6 @@
 			</li>
 			<li data-lang="en" id="91">
 				<a href="https://sixey.es/">sixey.es</a>
-				<a href="https://sixey.es/feed.xml" class="rss">rss</a>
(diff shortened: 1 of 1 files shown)


neauoire:
Thanks a lot for taking the time, I doubled check the links and they seem to be all down as you said :) cheers! There's a few website owners that I know personally and will reach out to ask if they want to add new feeds.

PR #806, opened 2023-06-27
Added my indict.us wiki page

nicksiv:
I've been following the webring for years. At last I have something ready to contribute. Hoping it's eligible since its one-page only! Thank you.


neauoire:
The web page is nearly 1mb to download, and I can't load it with my current internet connection. Could it broken down into pages?

nicksiv:
Yes but I need to change the way it exports. It's a tiddlywiki.
I forgot about the limitations. I'm sorry about that. You can cancel the pr
and I will get back with plain html pages.

Thank you!

PR #746, opened 2022-11-17
add benji.dog to webring

benjifs:
I've been wanting to join the webring for a bit but I hadn't been working on my website as much. I've been working on a few things recently and feel like its at a good place to submit. No JS and most things work as expected across all browsers tested (Firefox, IE, Netsurf).

webring icon is at the bottom of the homepage.

PR #807, opened 2023-07-04
add benji.dog to webring

benjifs:
A few months ago I submitted my site when it was lacking some content and organization. #746

I've since updated things as I'm back to actively posting now after a brief hiatus and thought it'd be a good time to resubmit.

JS only necessary for dark/light mode switch and easter eggs :). Tested on Firefox, IE, and Netsurf (`/watched` page is a bit off on Netsurf but actively looking into solutions).

webring icon is at the bottom of the homepage.


neauoire:
Looking great :) It's a big improvement on the last time I was here, welcome to the webring!

PR #818, opened 2023-08-08
Add notebox.fcjz.xyz

fcjz:
I put the icon at the bottom of the site.

You can see it no matter what.


neauoire:
Hi :)

I can't seem to load the website, it just loads and loads and eventually it fails, I managed to load up to 1.5mb, and nothing was showing up still. Also, at the moment, the webring needs you to have your own domain name.

PR #828, opened 2023-09-17
add monoxane.io

monoxane:
Hey, been following along for several years now and even started my own webring because of this, maybe it's time to attempt to join the real one :) 

Icon is located in the footer on the right.

I don't have a _lot_ of content right now but I'm actively working on adding more as I go and hope to fill it out a fair bit more in the near future. There's a few projects, some ideas I'm still working on, things some people would call blog posts, and other general digital garden nonsense. I do understand if this is deemed to be "not enough content" and will come back in a few months if that's the case, no dramas there.

It's all semantic HTML with minimal CSS and no JS (outside specific project pages that use it), and should be accessible in anything from Lynx or Mothra all the way to everyone's least favorite, Chrome.

### DIFF
--- a/index.html
+++ b/index.html
@@ -753,6 +753,9 @@
 				<a href="https://foreverliketh.is/blog/index.xml" class="rss">rss</a>
 				<img src="https://foreverliketh.is/docs/connectivism/home-page/buttons/foreverliketh.is.png"/>
 			</li>
+			<li data-lang="en" id="monoxane">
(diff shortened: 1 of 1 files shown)


neauoire:
Hi Oliver,

The website is pretty early in its development but there's enough content already and I'm certain it'll become something vast in time :) Lovely Helvetica Blanc wallpaper ;)

Welcome to the webring! See you on the fedi ^^

PR #832, opened 2023-10-13
Add ratfactor's website to the webring at their request

eli-oat:
Adding ratfactor to the webring at their request.


neauoire:
thanks for facilitating this :)

PR #834, opened 2023-10-15
Add hideout.ink

luisquid:
Link to webring can be found on the home page and on the footer of each post.


neauoire:
ss2
ss1
ss

The wiki pages look broken? Could you have a look :)

luisquid:
Hi! I am not sure what might be wrong. D:
I checked this with three different browsers and it all seems to work. 
It should look something like this: 
image
image
image

Would you mind sharing the browser you are using to see if I can see the issue?

Thanks!

luisquid:
Adding the last wiki pages I have: 
image
image

PR #839, opened 2023-11-02
Add parth.ninja to webring

partheus:
Hi!!!! I added the webring icon to the footer, so it will be on all pages at the bottom. There are a various pages that talk about my hobbies and work, and there is also a contact page at /hello. There are blog posts and other stuff as well.

Thank you and also thanks for maintaining this really fun webring!


neauoire:
Hey Parth! 

Your website is absolutely gorgeous :) Glad to be adding it to the webring, I love that the about page leads to all these different subsections :) Looking forward to following it grow!

PR #840, opened 2023-11-17
Add Autotel.co site to ring

autotel:
added my portfolio website, which I just tested to make sure that complies with what's required

link is at the bottom of index page.


neauoire:
I'm having a really hard time loading your website with our crappy internet connection, I'll have another look when we sail closer to the cell towers. How large is the index page just out of curiosity?

autotel:
12 mb :)

neauoire:
The design is too hard to use and the website is too slow to be a good candidate for the webring at this time, sorry.

PR #859, opened 2024-01-01
Add aksui.com

notji:
This webring has been one of my biggest sources of inspiration lately, leading me to try carve out my own home on the internet  again. I'd love to be a part of it, if you'd have me.

Webring link is in the footer of every page.

Thank you,
Benjamin


neauoire:
Hi Notji,

The website looks great! But it's too early, there needs to be 10 pages that are not meta pages or blog posts. Keep working on it and when it's further down the line, open a new PR :)

PR #862, opened 2024-01-05
add gehninfel.xyz to the list

hermanhel:
The webring icon is at the bottom of [gehninfel.xyz](https://gehninfel.xyz).

### DIFF
--- a/index.html
+++ b/index.html
@@ -778,6 +778,10 @@
 			<li data-lang="en" id="zinzy">
 				<a href="https://zinzy.website">Zinzy Waleson Geene</a>
 				<a href="https://www.zinzy.website/feed.xml" class="rss">rss</a>
+			</li>
+			<li data-lang="en" id="gehninfel">
+				<a href="https://gehninfel.xyz">Gehninfel</a>
+			</li>
 		</ol>
 		<footer>
 			<p class="readme">


neauoire:
I try to keep the webring as a place where people share some of their artistic projects as well, I have a hard time finding your projects in the zettle, none of the links works? Could it be that it requires javascript?

hermanhel:
>  I have a hard time finding your projects in the zettle

That is probably because I don't really have anything ready for show yet :cry:

 I do have some ideas expecting implementation in hold, but it's gonna be some time before I have them settled and hosted publicly available. Right now the website is more of zettelkasten note collection with some compilation on tutorials thingy. 

And I'm aware of the poor structure of the whole thing. a large index zettel seems a good idea at first but now more of an excuse of laziness :laughing:. I do plan to refactor it into one (or several) zettel that will make navigation and finding stuffs easier (such as less links/content on each page and more links and pages overall) in the following week

> none of the links works? Could it be that it requires javascript?

I am aware that some of the links do not work quite as expected, and I just fixed some this morning(reference and bibliography).

 I think other part of it may be fault of my emacs setting(I'm using org-mode to write all my documents and a system implemented in emacs-lisp to export org-mode to html), because some of them don't have enough information/have wrong information in raw html to start with. There are cases where I have links in org-mode but nothing in html. I am looking into it, though it may take some time before I figure them out.

And the javascript is a bit of a headache while I don't quite understand how it work(I just took a theme from [here](https://github.com/fniessen/org-html-themes) and it turns out to be js-heavy) I'm switching to a css-only theme that looks like [this](https://olmon.gitlab.io/org-themes/gray/gray.html) and there should be no js nor js related problem shortly after

neauoire:
The website is too early in its development to be a good candidate for the webring at this time, keep building it :)

PR #886, opened 2024-03-26
Add notes.zachmanson.com

zachpmanson:


neauoire:
Could you put the webring icon in the footer instead of putting it among the content?

PR #896, opened 2024-04-26
Revert to my original domain

r14c:
the tilde works better thematically and makes the url isomorphic with the gemini site.


neauoire:
I've merged it, but just a head's up:

ss

r14c:
@neauoire thanks! The server that handles the renewals has been down for maintenance. It'll be back in working order soon!

PR #914, opened 2024-06-13
change button source for zvava.org

mininmobile:
`/images/button.gif` → `/images/buttons/zvava.org.png`


neauoire:
Updated :) thanks!

PR #920, opened 2024-07-04
Add breckyunits.com

breck7:
Webring location: https://breckyunits.com

### DIFF
--- a/index.html
+++ b/index.html
@@ -780,6 +780,10 @@
 				<a href="https://mattmcadams.com">Matt McAdams</a>
 				<a href="https://www.mattmcadams.com/feed/feed.xml" class="rss">rss</a>
 			</li>
+			<li data-lang="en" id="breck">
+				<a href="https://breckyunits.com">Breck Yunits</a>
+				<a href="https://breckyunits.com/feed.xml" class="rss">rss</a>
+			</li>
 			<li data-lang="en" id="lukaprincic">
 				<a href="https://luka.princic.studio">Luka Prinčič</a>
 			</li>


neauoire:
The page is too large for me to load at the moment, it's not a good candidate if I cannot review the site.

breck7:
@neauoire what kind of computer are you using?

There's also a completely text version of every page. Just swap *.html with *.txt

https://breckyunits.com/index.txt

breck7:
@neauoire can you provide some more details? this is the first complaint Ive ever gotten about the page being too large. Very curious to know whats going on.

breck7:
@neauoire seriously the whole site shows 2 decades of blog posts. There is no script files, no analytics. There's a completely plain text version. The source code for every page is included. It has a unique homepage where you can see ~20 years of my thinking zoomed out (but you can zoom in as well).

Just because it is different, doesn't mean it is bad. Please give it some thought.

neauoire:
Hi,

Yes sorry, I only just got back within reliable connectivity yesterday and I'll have a second look later today :)

breck7:
Thank you @neauoire ! I really appreciate it. Perhaps there's a way I can better serve more limited devices. I do use CSS3 columns, the layout computations, might not be well optimized in older browsers. Or maybe I can further reduce image sizes. 

I appreciate the feedback.

neauoire:
> @neauoire what kind of computer are you using?

It's a x230 thinkpad, running on satellite internet. Your website downloads some of the html and then just halts, I think my connection gives up before finishing the download and so I have an incomplete file, but I'm not sure, I haven't had that issue before except with very large pages. I use Netsurf as browser, so I don't see 3 layouts, just one and that's alright.

I'm wondering if maybe I haven't found a glitch in netsurf itself..

breck7:
Ok, thanks for the info, I'll put it on my todo list to test with Netsurf.

PR #926, opened 2024-07-20
Hardcode <img> dimensions to prevent reflows

noelleleigh:
When toggling the visibility of the icons, the list items shift around a bunch as all the images are loaded. By specifying the `width` and `height` of the images, the browser can adjust the layout instantly before they are even loaded, eliminating the reflows.

Also set `overflow: hidden` on the images so that the alt text that appears when the image won't load doesn't trigger a reflow.

Also updated the examples in the README to include `width`, `height`, and `alt`.


neauoire:
Good idea! Thanks : )

PR #933, opened 2024-08-29
looking to join the webring

eliobrid:
https://eliobrid.neocities.org/webrings


neauoire:
I currently don't add websites that don't have their own url, sorry.

eliobrid:
ah ok, that makes seance
i'll look into getting one eventually
i'll be back in time i think

PR #938, opened 2024-09-19
add ephemeral.cx

shanet:
Hello! May I join with https://ephemeral.cx please? 😊

> Submit a Pull Request with the location of the webring icon on your site. 

The webring icon is located in the navigation on the index page and also on the /about page.

### DIFF
--- a/index.html
+++ b/index.html
@@ -856,7 +856,11 @@
 			</li>
       <li data-lang="en" id="laeastra">
 				<a href="https://laeastra.com">laeastra</a>
-			</li>	
+			</li>
+			<li data-lang="en" id="ephemeral">
+				<a href="https://ephemeral.cx">ephemeral.cx</a>
+				<a href="https://ephemeral.cx/feed.xml" class="rss">rss</a>
+			</li>
 		</ol>
 		<footer>
 			<p class="readme">


neauoire:
Hi Shane,

Your website looks nice but it's sooooo heavy, it kicks my connection offline everytime I try to load any of the pages. I can't quite see why that happens, but I think maybe the images are not optimized for web or maybe there's too much embed content. Try optimizing things a bit if you can and open another PR, I'd love to have a proper look around if you do!

shanet:
Hmm, I wonder what's going on then because I don't see that on my end. My website is all static HTML with almost all the pages under 1mb of content including the images. I've tried to keep it as minimal as possible.

The only heavy pages are the photography albums which is intentional since the point is to showcase high resolution images (and even those are downsized to smaller sizes than the originals too for the sake of load times with the preload set to be after the thumbnails to prioritize the initial page load). Maybe your browser is trying to pre-load those before the pages are requested?

For the index page, about, and blog page I see everything measured in kilobytes. Even the aviation videos page is ~2MB until a video is played:

screenshot_2024-10-02-105833
screenshot_2024-10-02-105854
screenshot_2024-10-02-105755
screenshot_2024-10-02-105814

PR #940, opened 2024-09-25
Add www.sdavidprince.space

sdavidprince:
Add my  Personal site www.sdavidprince.space.

The Linkback is added at the footer of the site homepage.
Hope this gets into the ring


neauoire:
ss

Sorry, I don't accept websites that rely on modern web browsers to work properly.

sdavidprince:
💁🏾‍♂️ I actually only runs on html css and js. Mxm only tested on opera. What are your speculations

PR #942, opened 2024-09-30
Add breckyunits.com

breck7:
For slow browsers, this site is also available as TXT: https://breckyunits.com/index.txt

Webring location: https://breckyunits.com/

### DIFF
--- a/index.html
+++ b/index.html
@@ -780,6 +780,10 @@
 				<a href="https://mattmcadams.com">Matt McAdams</a>
 				<a href="https://www.mattmcadams.com/feed/feed.xml" class="rss">rss</a>
 			</li>
+			<li data-lang="en" id="breck">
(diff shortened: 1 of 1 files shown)


neauoire:
Sorry the website is [a blog](https://github.com/XXIIVV/webring?tab=readme-ov-file#webring-criteria), and is missing an about page, and the webring has a focus on arts and I cannot find content of that type in the blog posts.

neauoire:
Em's LowTech webring is excellent.
https://emreed.net/LowTech_Directory

ckiee:
oh wow your layout is so dense -- I would never do that but I kinda love it

breck7:
thanks @ckiee! I find it helpful when rereading my own stuff "syntopically" (https://en.wikipedia.org/wiki/How_to_Read_a_Book).  there's also now a version without the full previews https://breckyunits.com/short.html

PR #943, opened 2024-10-05
Added wiki.78bpm.com

century:
Added wiki.78bpm.com

You can find the webring icon at the bottom of every page. It is a collection of resources I have used as a basis for my own work, and it generated using a DIY site generator.

### DIFF
--- a/index.html
+++ b/index.html
@@ -868,6 +868,11 @@
 					height="31"
 				>
 			</li>
+			<li
+				data-lang="en"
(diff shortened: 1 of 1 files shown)


neauoire:
Good content and all but it's a bit too minimal at the moment, I try to gather on the webring websites with at least a handful of pictures and a bit of inclination toward design and arts. Feel free to open another PR if you decide to push the design further.

PR #944, opened 2024-10-13
add leetusman to webring list

lee2sman:
The merveilles webring icon can be found in the footer navbar of my [index page](https://leetusman.com) and on every other page on my site.


neauoire:
Gorgeous website, which I was already familiar with ^__^ welcome to the webring Lee!

PR #945, opened 2024-10-17
Add triapul.cz

caffo:
Adds triapul.cz (from our lovely) [@prahou](https://merveilles.town/@prahou)


neauoire:
Hell yes, thanks for doing the relaying of PR from prahou

PR #946, opened 2024-10-22
Add cujo.casa

cujohub:
Hello, I'd love to have my website added to the webring. The link to the directory is in the footer of my homepage ([cujo.casa](https://cujo.casa)). Thanks in advance!

### DIFF
--- a/index.html
+++ b/index.html
@@ -868,6 +868,10 @@
 					height="31"
 				>
 			</li>
+			<li data-lang="en pt" id="cujo">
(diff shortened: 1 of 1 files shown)


neauoire:
Hey ditocujo,

Nice website, but it is missing an about page with contact info, and all the projects are offsite links, none is documented on the website itself. In any case, it's currently missing creative works, I try to gather on the webring a collection of links that showcase an interest in arts somewhat, whether it be illustration or design, but it has to be more than code and scripts.

PR #954, opened 2024-11-09
Add arne.xyz

akamola:
Hi! I would like to join this webring.

Website: https://arne.xyz

The icon of the webring is in the footer on the left hand side an available on every page.

Thanks in advance!


neauoire:
Very nice website (and zine!) Arne.
Welcome to the webring : )

PR #955, opened 2024-11-11
Update index.html with 88x31 + url/name change

lukaprincic:
added 88x31 button and changed the url (the old works and forwards to the new one) and changed the name (but not 'id', yet)


neauoire:
nice! thanks :  )

PR #963, opened 2024-11-25
li shouldnt split between columns

Rezmason:
li (list item) tags necessarily have a CSS display of "list-item", which may cause their contents to spill across the ends of columns. By wrapping their contents with a span tag, the contents are fit into an "inline" element and stick together.


Rezmason:
Wrong base branch. Let's try this again

PR #964, opened 2024-11-26
Wrapping li contents in span tags (attempt 2)

Rezmason:
li (list item) tags necessarily have a CSS display of "list-item", which may cause their contents to spill across the ends of columns. By wrapping their contents with a span tag, the contents are fit into an "inline" element and stick together.

### DIFF
--- a/index.html
+++ b/index.html
@@ -9,798 +9,798 @@
 	</head>
 	<body id="twtxt">
 		<ol id="icons">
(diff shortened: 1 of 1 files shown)


Rezmason:
I haven't addressed you folks as Sea Peoples in several years 😄

neauoire:
It adds too much junk in the file I think :/ It's fine if it overflows.

PR #976, opened 2025-01-03
Add ike.io

ike:
This adds my site, `ike.io`, to the webring. :)


neauoire:
Hey Isaac, 

Your website is nice but too much like a blog at the moment, I try to collect websites in the webring that breaks off from the single timeline format a bit. If you do decide to grow it more like a digital garden, please open a new PR :)

PR #1006, opened 2025-04-09
Add caveli.us

walidcavelius:
Add caveli.us
icon is included in the navbar on the left in every page


walidcavelius:
i think i don't have enough content actually, i'll do a PR when i feel it's ready!

PR #1009, opened 2025-04-12
Add icon for manifoldslug

manifoldslug:
Made an 88x31 button for the wreckage, and I'd like to add it to the fold.


wakest:
@manifoldslug I can't get this to load, is your site down?

PR #1015, opened 2025-04-29
Add hmd-tld.xyz

hmd-tld:
Hello
I want to add my website to the webring. The icon is at https://hmd-tld.xyz/acercade.html
Greetings


neauoire:
The site is off to a good start but it's too early in its development to join the webring, send another PR when it's further down the line :)

PR #1018, opened 2025-05-06
Add hamishjohnson to index.html

Ham0osh:
Added logo and hyperlink to footer. Plan to elaborate in a way that lets users travel through the ring.

### DIFF
--- a/index.html
+++ b/index.html
@@ -897,6 +897,9 @@
 				<a href="https://eevie.dev">eevie.dev</a>
 				<img loading="lazy" src="https://eevie.dev/img/button.png" alt="eevie" width="88" height="31">
 			</li>
+			<li data-lang="en" id="ham">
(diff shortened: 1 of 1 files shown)


neauoire:
I'm not sure why but your website is loading really slow for me, I'll try again when my connection is faster. Are the images optimized for web? The project page only managed to load a few things before giving up.

Ham0osh:
Not sure. They load a little slow sometimes but I havent had any issues like that. Currently most project cards use smaller webp but I'll reduce them to more of a thumbnail size and follow up. Similarly the home page uses src set with sizes that match lighthouse recommendations but I'll bring them down a bit.

neauoire:
Your website is just too heavy I think, clicking on projects makes my browser spin without displaying anything. I can see a ton of requests, but it's just too much for our crappy connection. Sorry.

PR #863, opened 2024-01-07
Add akavel.com

akavel:
The icon should hopefully be visible at the bottom-right of [the index page](https://akavel.com)(s).

FWIW, the sources are currently at: https://github.com/akavel/garden

PR #1021, opened 2025-05-06
Add akavel

akavel:
[take 2](https://github.com/XXIIVV/webring/pull/863), a year later :)

the ring icon is at the bottom of the [main page](https://akavel.com), on the right


neauoire:
The site has growing nicely since I last checked.

I usually don't accept websites to the webring that documents projects on third parties(github, itchio, etc..). 
But I think [Projects](https://akavel.com/projects) should be linked from up top, and if you ever decide to document the projects on your personal website itself, it would be even nicer :)

In any case, welcome to the webring!

PR #1028, opened 2025-05-25
Idiomdrottning

snan:
The webring link is in three places, at the bottom right of the
/gallery and /texts and on the front page / . I removed the target
blank and made it just normal links. On the /texts page I moved the
word "webring" from the alt text to the a element because it's a
/texts page.

Also my PR didn' have this stuff first because github confusion. it was in the commit message. I'm more used to using git without hub

### DIFF
--- a/index.html
+++ b/index.html
@@ -912,6 +912,12 @@
 				<a href="https://stage7.net/rss/rss.xml" class="rss">rss</a>
 				<img loading="lazy" src="https://stage7.net/links/stage7.png" alt="stage7" width="88" height="31">
 			</li>
+			<li data-lang="en" id="Idiomdrottning">
+				<a href="https://idiomdrottning.org">Idiomdrottning</a>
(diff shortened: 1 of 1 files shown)


snan:
From the gallery the first icon with my thistle logo leads to the /about page. But the gallery and the front page and also /texts all have the webring links.

For the first ten years of the web page's existence, what's now the /about page was the front page! Then from 2020 to 2024, what's now the /gallery was the front page and the first icon led to the old front page (renamed /about). And after a few years I added the /texts page which has never been the front page, only linked from the gallery. Then last year I made a new front page with Misa saying that she doesn't use the internet and it linked to texts and the old front page (now renamed /gallery )  but not to /about.

I wanna dive into the whole 100rabbits verse a little bit more because I like that kana trainer and I loved loved loved Busy Being Nothing. and of course merveilles!

I'm so grateful to be accepted to this awesome webring!

PR #1032, opened 2025-06-02
Cleanup index

snan:
Removed stray whitespace and tags.

Now, if the spaces around itwont.work were intentional, please restore
them. And if the italics around embyr.sh were intentional, please
restore them and also add to mine. italics the best


snan:
Also is it intentional that a:visited does not get {curlies around them}? I spent some time trying to debug why some links didn't have the curlies and figured out that it was the visited ones.

neauoire:
yeah, that's by design : )

PR #1034, opened 2025-06-09
joining the ring

Cosmhyccc:
Hello
Added my site https://saif.build to the webring.
Webring icon is up at https://saif.build/button.gif Would be honored to join the crew


neauoire:
All projects are documented offsite, sorry, the site does not qualify.

neauoire:
Honestly, since the webring is meant to showcase the craft of artists and developers, and that these are all AI projects I don't think it's also a good fit in terms on content either.

PR #1041, opened 2025-07-08
add nnix.com to ring list

davidemerson:
- Webring icon on every footer (https://nnix.com has it), footer element is common so it'll be every page when rendered.
- Migrated the rest of my old gemini capsule over, the site is a lot richer now.


neauoire:
I looove what you did with the place : ) Much more interesting, and the splash is a nice touch.
Welcome to the webring!

PR #1044, opened 2025-08-07
Fix my site

dal3m:
I am very sorry to create another pull request just after being added! But I screwed up, when first committing I used to use .png now i almost exclusively use .jpg for most images on my site, rendering my banner broken. Also added my blog's rss feed to opml file.

Once again sorry for the troubles and thanks so much for being added!

Cheers,
- dalem.


neauoire:
I don't mind merging it, but png is a better format for small graphics like this as they are lossless. I suggest that you keep png for icons, and interface details, and of still pixel graphics. If you'd still like for me to merge this one, let me know : )

PR #1048, opened 2025-08-22
Add dill64

Nightsquid7:
Hi,

I would like to add my site [https://www.dill64.com](https://www.dill64.com) to the webring.

I have the webring logo on the footer of my home page, and at the top of most other pages.

You can find my about page [https://www.dill64.com/about.html](https://www.dill64.com/about.html) clicking the robot icon on the home page

Thanks!


neauoire:
Doesn't work without javascript, sorry.

PR #1049, opened 2025-08-25
Update domain for wolf.md to wol.fm

wolfmd:
In an act of impulsive vanity, I purchased a domain name that is three letters shorter and less suggestive of me having medical credentials. Updates the link and rss feed for wolfmd.me to wol.fm.

I intend on maintaining both domains for the foreseeable future but I feel obligated to make use of the cooler (it is cooler, right?) domain whenever possible

An update to the button will be made in conjunction to this PR

Holler at me if I need to make any adjustments to this commit or otherwise and thank you for helping build this awesome community

### DIFF
--- a/index.html
+++ b/index.html
@@ -389,9 +389,9 @@
 				<a href="https://notebook.wesleyac.com/atom.xml" class="rss">rss</a>
 			</li>
 			<li data-lang="en" id="141">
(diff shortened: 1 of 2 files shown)


neauoire:
Haha, that's totally fine ^__^ thanks for the head's up

PR #1053, opened 2025-09-21
request to add xcv.gr

glyras:
Hello! 
My PR to be added to the webring. Icon is located on footer!

Used id="95" because it was available and appropriate for 'xcv' :)

Fair winds!


neauoire:
Hi :)

Your website has too few pages to qualify for the webring, but more generally, the webring is aimed for artists and designers with websites that needs to demonstrate a bit of creative works beyond computer things like illustration, photography, sculpture, music, etc...

PR #1065, opened 2025-11-20
added artart.ch

LiohMoeller:
Webring Link can be found at the bottom of the page https://artart.ch


neauoire:
Unfortunately, the website doesn't work without javascript so it does not qualify for the webring at this time. If you get around to make it work without it, let me know!

PR #1068, opened 2025-12-11
Retsagames.com

RetsaGames:
Hello there! i'm Retsa, a game developer/programmer/artist and i wanted to add mi website [retsagames.com](https://retsagames.com/) to the webring.
It's mostly a portfolio containing my games, some of which can be played on the site itself. It contains some journals and miscellaneous stuff.
The site is lightweight & simple, uses no framework & javascript is only used for the web games.
The index page has an "about" section.


RetsaGames:
true, my intent is to have more content pages apart from a portfolio of my creations, but that aspect is currently lacking.
Will send a push request again when the site has more of my thoughts/journals in.

PR #1072, opened 2025-12-21
Update index.html

stephanelambion:
Hiya! Location of the link is in "friends with websites" on the homepage :)


neauoire:
Lemme check back from time to time :) we'll keep the PR open, you can take off the webring icon if you like in the meantime.

PR #1107, opened 2026-05-18
Add vfnvfn.net

fkinos:
hi, i would love to share the webring with so many awesome websites, just not sure if my website currently fits in

feel free to reject if it doesn't!


neauoire:
It's off to a good start, keep working on it and when it's a little further down the line it'll make a great addition to the ring :)

PR #1109, opened 2026-05-22
add quasivoid.net to webring

f-s-sparverius:
hi,

the icon is in the footer of every page, left of the 88x31 button collection.

thanks


neauoire:
Nice capsule : ) Welcome to the webring!

PR #1117, opened 2026-06-29
Add The Winter Throat to ring

Anahkiasen:
The webring link is accessible from the [Webrings page](https://autopergamene.eu/webrings) linked in the footer on every page. Hope I did everything correctly 🤞

### DIFF
--- a/index.html
+++ b/index.html
@@ -977,6 +977,11 @@
 				<a href="https://xo.codes">xo</a>
 				<img loading="lazy" src="https://xo.codes/img/banner.gif" alt="xo" width="88" height="31">
 			</li>
+			<li data-lang="en" id="anahkiasen">
+				<a href="https://autopergamene.eu">anahkiasen</a>
+				<a href="https://autopergamene.eu/api/feed" class="rss">rss</a>
+				<img loading="lazy" src="https://autopergamene.eu/button.png" alt="anahkiasen" width="88" height="31">
+			</li>
 		</ol>
 		<footer>
 			<p class="readme">


neauoire:
Unfortunately, your website somehow makes my browser(Firefox SRE) crash? The page becomes unresponsive and halts. I'm unable to see the website, I can see it's starting to load a lot of assets(A LOT) in the debugger, but it's too much for my old laptop I think X) If you simplify it enough so that an old computer like mine can load it, please do open a new PR ^__^

Anahkiasen:
I'm sorry I was maybe working on it at the moment or one of the webring widgets was messing with things perhaps (I've had issues with those and had to remove some)? I'm myself on Firefox and have worked to make it light and fast, there's even SSR so it works without JS, so I do want to make it work. The assets count is because they are lazy loaded to load less and they're HTTP/2 which should make that efficient even on less performant laptops :s Was it on the webrings page or all of them? The only "animated" thing is the cursor follower but that's rather basic and can be disabled in the top right, not sure if thats what caused it :s

neauoire:
https://github.com/user-attachments/assets/8bc14f90-91d0-4bc6-b742-fc0766a8f4fb

I can help you improve it yes, gladly :) There's too many issues with firefox to pin them down, I don't really understand why, but moving my cursor requests full screen redraw at 60fps, pinning my CPU. Even clicking on the widget at the top right is very difficult because it's lagging as it expands.

Regardless of the CSS issues, I noticed that the content of every page exists twice in the body of each page? I think that might be a hint of the issue.

Anahkiasen:
Okay that's weird that's not at all what the website looks like, this is basically the unstyled version. The two bodies are the desktop and mobile layouts and they hide responsively normally but same, it's in the css that apparently is not loading :/ I'll see what I can find in my logs and what I can improve but I'm not doing anything out of norm  this is supposed to be relatively light so there definitely is something wrong 😅 this is what it should look like. 

image

Maybe my website doesn't handle user styles well too. The request on cursor move is a requestAnimationFrame for the dumb cursor follower it's not a big bad thing normally it's meant the be the most efficient way too :/ I'll check that too 

I'll try to restore my honor. Sorry for the time spent

Anahkiasen:
I'm sorry maybe I misunderstood this part of the description:

> If your website requires Javascript/CSS3 to display the majority of its content or to navigate, it will be rejected

I had focused on the JS part of course in the idea of a website that doesn't require fancy shit to work (for example most animations are plain transitions) but ultimately I want to make a fun quirky cool website that hooks people in, even if I work to make it printsheet compatible most of my art is visuals, I'm not sure I can make it a cool experience without CSS or loading images, it's not aiming to be a resilient/lowtech just to be my inspiring website as an artist and developer. So that part of the requirement it fits but not the technical part I think, so I'm not so sure this fits in the ring now :s 

I admire your own website for the lowtech aspect and how it still manages to be beautiful and fun, so I'm not saying these concepts are incompatible at all to be clear! just this website has 2 decades and its always went a different direction, maybe its not reconcilable >_< sorry again

PR #1120, opened 2026-08-03
add omz13.com

omz13:
Please add my personal website (https://www.omz13.com) to the webring.

The location of the webring icon is in the footer.


neauoire:
The website is off to a nice start but it has too little content in its current state to qualify for the web ring :) Keep building it up!

PR #1123, opened 2026-08-09
remove si3t.ch from index.html

B4rb3rouss:
SInce I won't keep the link to the webring on my website, I should be removed.


neauoire:
Thanks for the head's up :)

PR #1124, opened 2026-08-10
add merc.guru to the webring :)

foobar-p5:
https://merc.guru
the icon is always visible in the footer


neauoire:
awyis, always love me a new werc site on the ring! welcome :)

PR #1128, opened 2026-08-16
add dogthoughts.link

kaybarkbark:
Webring icon is in the footer, in the blue Webring box

You can contact me at kay@dogthoughts.link if you have questions, or just reply to this PR. I have a fair amount of projects listed, I think they qualify as portfolio pages instead of blog posts. If you disagree, that is ok! I understand. :) Thank you for maintaining a webring!


neauoire:
Cool page, nice hardware hacks :) Welcome to the webring!

